PR #2, opened 2017-11-28
Would be nice to have this in repository

mkudrnac:
+ AM2302 dependency

### DIFF
--- a/sensors/DHT22.py
+++ b/sensors/DHT22.py
@@ -1,0 +1,283 @@
+#!/usr/bin/env python
+
+# 2014-07-11 DHT22.py
+
+import time
+import atexit
+
+import pigpio
+
+class sensor:
+   """
+   A class to read relative humidity and temperature from the
+   DHT22 sensor.  The sensor is also known as the AM2302.
+
+   The sensor can be powered from the Pi 3V3 or the Pi 5V rail.
+
+   Powering from the 3V3 rail is simpler and safer.  You may need
+   to power from 5V if the sensor is connected via a long cable.
+
+   For 3V3 operation connect pin 1 to 3V3 and pin 4 to ground.
+
+   Connect pin 2 to a gpio.
+
+   For 5V operation connect pin 1 to 5V and pin 4 to ground.
+
+   The following pin 2 connection works for me.  Use at YOUR OWN RISK.
+
+   5V--5K_resistor--+--10K_resistor--Ground
+                    |
+   DHT22 pin 2 -----+
+                    |
+   gpio ------------+
+   """
+
+   def __init__(self, pi, gpio, LED=None, power=None):
+      """
+      Instantiate with the Pi and gpio to which the DHT22 output
+      pin is connected.
+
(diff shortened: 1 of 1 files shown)


mkudrnac:
Well this not pip3 module and I did not find any repo where this is in

Solutions:
1) make separate repo for this and use submodule
2) same as 1) but not git submodule -> link this in comment like https://github.com/maxlklaxl/python-tsl2591 and provide setup.py for easy import 
3) create pip3 module for this -> install pip3 install dht-22 or whatever name

Personally I would prefer 2 or 3 because it easier for new users. git submodules are not so user friendly because git clone by default is not downloading it

What do you think? :)

ikalchev:
I like 2 very much. How do you imagine changing setup.py? Did you mean using "dependency_links" for packages that are not in pip, similar to [this](https://stackoverflow.com/questions/26061610/pip-install-dependency-links)?

I will definitely look into it, but any experience you can share is welcome, as I haven't done something similar.

mkudrnac:
with option 2, I mean that create git repository for example dht22 and create setup.py like HAP-python has, so user can 
```git clone <repo>``` 
 ```python3 setup.py install``` 
and than in AM2302 accessory just use import dht22.DHT22 as DHT22
This looks much easier than download file from some forum and copy it somewhere

Example of setup.py
```
from setuptools import setup

setup(name='dht22',
      version='0.0.1',
      url='http://github.com/ikalchev/dht22',
      author='<author>',
      author_email='<email>',
      description='Community-coded Python module for dht22 sensor. Use it at your own risk.',
      packages=['dht22'],
      long_description=open('README.md').read(),
      requires=['python (>= 2.7)'],
      zip_safe=False)
```

Or create HAP-python-sensors so we can add more than one :-)
import sensors.DHT22 as DHT22
import sensors.tsl2591 as tsl2591
etc.

What do you think?

ikalchev:
I like the option with standalone repos and links to them in each accessory. It seems more convenient, because if you use, say, Adafruit's sensor libs, you wouldn't add them in HAP-python, but put a link to the lib's GitHub repo. Agree?

mkudrnac:
Yes, I like it too :-) what about name HAP-python-sensors?

ikalchev:
Really sorry for replying so slow for this, I expect to be more prompt now the weekend is gone.

Could you clarify on why we need a specific name? The way I imagine is that, e.g. I write some sensor code and put it a repo e.g. `ikalchev/FooSensor.py` (with setup.py in there) and then write a `pyhap/accessories/Foo.py` in which I put a link to the repo so that everyone interested can **git-clone-python-setup-install** it. Did you have the same in mind or I misunderstood?

mkudrnac:
That's what I meant :-)

mkudrnac:
I thought that you could put all sensors in one repo with all sensors but separate is better

ikalchev:
Okay agreed then. I am closing this now. Let me know if you put the AM2302 code in a repo so that we can add a link in the respective accessory :)

mkudrnac:
👍

PR #8, opened 2018-01-21
OptionalCharacteristics is missing

xtqxk:
add OptionalCharacteristics for service

### DIFF
--- a/pyhap/loader.py
+++ b/pyhap/loader.py
@@ -50,6 +50,8 @@ class ServiceLoader(TypeLoader):
         s = Service(type_id, name)
         chars = [self.char_loader.get(c) for c in serv_info["RequiredCharacteristics"]]
         s.add_characteristic(*chars)
+        opt_chars = [self.char_loader.get(c) for c in serv_info["OptionalCharacteristics"]]
+        s.add_opt_characteristic(*opt_chars)
         return s
 
 


ikalchev:
This is by design and is documented in https://github.com/ikalchev/HAP-python/blob/c1ceb50c155720a548e068415fc93e3203d7ab65/pyhap/service.py#L15

My reasoning was that the chars are called optional for a reason - not every service of this type has these characteristics. Therefore, if we add the chars to the service, they will be displayed to the iOS Home app, but the user (of HAP-python) may not intend to use them, so they will just clutter the UI.

Explicitly adding optional characteristics is shown in [the fake fan accessory](pyhap/accessories/FakeFan.py). If you have different thoughts on the above, let me know :)

xtqxk:
I used Http:
eg:
http = Http(("", 51800), ["TemperatureSensor"], "HTTP bridge",
            aid=STANDALONE_AID, mac=util.generate_mac(), pincode=b"203-23-999")

then post data:
{ "TemperatureSensor" : {
     "CurrentTemperature" : 20,
     "Name":"xxx"
 }
}

the attr Name will throw exception.

And how to implement mulit TemperatureSensor to one service use http bridge?

ikalchev:
> How to implement multi TemperatureSensor?

You can just pass more TemperateSensors in the list e.g. `[“TemperatureSensor”, “TemperatureSensor“]

As for the Name: I really haven’t thought about optional chars in Http Accessory, but it is definitely good to add them. Otherwise there is no way to distinguish the above multiple services (except spawning multiple Http accessories, which is an overkill).

I will open a bug to track this and will try to push a solution in the following days. I will keep you posted.

Thanks for noticing :)

Best regards,
Ivan

Issue #26, opened 2018-02-15
Allow characteristic properties like minValue to be overridden.


ikalchev:
It should be possible to configure the properties of a characteristic to align with the functionality of the device and the iOS UI. For example, say your window covering can go up only to 50% open and that it is either closed or open. HAP specifies the default values for the coverings target position as maxValue=100 and minStep=0.1. This means iOS clients can slide at 0.1 step all the way to 100%, which you must map to all-up or all-down in your code.

Instead, you can do:
```python
service = loader.get_serv_loader().get("WindowCovering")
pos = service.get_characteristic("TargetPosition")
pos.properties["maxValue"] = 50
pos.properties["minStep"] = 50
pos.hap_template = pos._create_hap_template()
```

Now iOS correctly allows you to only fully go up or down and the max display value is 50%. The hacky thing about the above is the last line and this is the topic of this issue - overriding properties should be more "native".

helmsman35:
Cool! This is important to properly configure the HomeKit accessories.

I guess a Pos.set_property(“maxValue”, 50) helper methode would be possible. Or even also support a dict as parameter to set multiple properties at once.  The call to _create_hap_template would go in there. I have not looked info the code of that method so I assume here that multiple calls to that method should be possible.

ikalchev:
Yes, correct. I will see which option is the easiest for clients. The whole problems comes from the fact that this hap template is a kind of skeleton dict which is appended to and sent every time to iOS clients. So to avoid computing this on every request, I cache the template, so property changes get left out.

btw another option would be use the dict interface all together and allow only properties defined by HAP:
```python
char["maxValue"] = 100  # ok
char["minStep"] = 2  # ok
char["foo"] = 2  # key error
```

helmsman35:
we must be careful not to do things without thinking.
this is only needed at startup and not at runtime. Setting this in the __init__ of the Accessory

        self.CurrentPosition = self.get_service("WindowCovering").get_characteristic("CurrentPosition")
        self.CurrentPosition.properties["minStep"] = 100

I guess it all depends on when self._set_services is called

ikalchev:
This is one option, but then you need to trigger the create_hap_template or it needs to be generated every time.

[...]

PR #28, opened 2018-02-15
Bugfix: Changed CurrentTemperature min/max values

cdce8p:
I noticed that the min value for `CurrentTemperature` was set to `0` which doesn't account for negative temperatures.

I really like your project. After I saw the discussion in the `HomeAssistant` Forum about implementing `Homekit` support, I tried a few things with your project myself 👍

### DIFF
--- a/pyhap/resources/characteristics.json
+++ b/pyhap/resources/characteristics.json
@@ -303,9 +303,9 @@
          "ev"
       ],
       "unit": "celsius",
-      "minValue": 0,
+      "minValue": -273.1,
       "minStep": 0.1,
-      "maxValue": 100
+      "maxValue": 1000
    },
    "CurrentTime": {
       "Format": "string",


ikalchev:
Hey thanks for contributing! I am happy you liked the project, keep posting issues like these.

About the issue - interesting enough, the CurrentTemperature definition is taken as-is from Apple's HomeKit Dev Kit, i.e. Apple is using 0-100 range as the default as well. But your change makes sense - who wouldn't want to support negative temperatures by default?

What worries me is that as soon as I re-parse the Dev Kit types to include the latest new Chars and Services (there are a few), your change will be "rolled back" (see `gen_hap_types.py` at the root of the project). The solution I was thinking about is to have a kind of `overrides.json` file, which will contain "patches" for Apple's default min/max values, etc. This file will be integrated as a last step in `gen_hap_types.py`. Thoughts?

In any case, this is good to go, merging!

PS Very nice touch with the absolute zero.

ikalchev:
FYI #26

cdce8p:
As long as Apple doesn't change the Dev Kit, this might be a good idea. I just started working with HomeKit, so I know to little to give actual advice.

An issue that might arise here is: What about Fahrenheit values? I guess you have to update the unit and consequently the values (Fahrenheit starts at: −459.67°) for it as well.

PR #30, opened 2018-02-16
Issue #24 - Initial X-HM:// URI and QR Setup Code work

jslay88:
For Issue #24 

Needs further discussion, such as how to make the bitwise ops more dynamic. 

My lack of proper Apple Documentation limits this for me. :)

PR #31, opened 2018-02-17
QR Code generation and dynamic pin/setup codes.

schinckel:
I'm not necessarily expecting this one to be merged in favour of https://github.com/ikalchev/HAP-python/pull/30, but just putting it here so we can compare the way I've done it compared to @jslay88.


schinckel:
I'm going to move the dynamic codes into a seperate PR.

schinckel:
No biggie: I'm not sure if I'd actually pushed it or not, to be honest.

PR #33, opened 2018-02-17
Add accessory to control Mac display status.

schinckel:


pyhap/accessories/DisplaySwitch.py
```diff
@@ -0,0 +1,51 @@
...
+        self.display.setter_callback = self.set_display
+
+    def _set_services(self):
+        super(DisplaySwitch, self)._set_services()
+        self.add_service(
+            loader.get_serv_loader().get("Switch"))
+
+    def run(self):
+        while not self.run_sentinel.wait(1):
+            # We can't just use .set_value(state), because that will
```
ikalchev: Hm, this is the second time I see a "request" for `set_value` to have something like `should_callback=True`. Do you think this would make sense?

pyhap/accessories/DisplaySwitch.py
```diff
@@ -0,0 +1,51 @@
+# An Accessory for viewing/controlling the status of a Mac display.
```
ikalchev: nit: could you please change this to a doc string

pyhap/accessories/DisplaySwitch.py
schinckel: Quite possibly.

ikalchev: Okay thanks I will open an issue to track this.

PR #34, opened 2018-02-17
Dynamic codes

schinckel:
Generate random `pincode` and `setup_id` each time the process is restarted (unless explicitly passed in).

This is in line with the recommendations from the Apple HAP documentation: https://developer.apple.com/support/homekit-accessory-protocol/

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -1,4 +1,3 @@
-import uuid
 import threading
 import logging
 import itertools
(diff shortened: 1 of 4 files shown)


pyhap/accessory.py
```diff
@@ -142,7 +141,8 @@ def __init__(self, display_name, aid=None, mac=None, pincode=None,
         self.mac = mac
         self.config_version = 2
         self.reachable = True
-        self.pincode = pincode
+        self._pincode = pincode
+        self._setup_id = setup_id
```
ikalchev: I would `setup_id = setup_id or self.generate_setup_id()`. Same for pincode.

Issue #35, opened 2018-02-17
Garage Door Opener


jslay88:
Trying to get this working, and it works for the HomeKit requests, but not when the door state changes outside of HomeKit (i.e. when I open the garage with my car remote, instead of HomeKit).

Would like some help troubleshooting this. I am getting the opposite transition state than you would expect, only when the door state changes outside of a HomeKit request from an iOS device.

You can easily simulate this on a RPi, by shorting out a GPIO pin to ground for a `CLOSED` state, and removing from ground for an `OPEN` state. On the iOS, device you can request open, then simulate open, and so forth.

To repro the issue I am having, simply simulate the opposite state of the door, without requesting it from HAP-python.

Also, would like some clarification on how to properly set the initial state of `CurrentDoorState`, as I was seeing some weird things there too, like starting out as `Opening`, even though it was clearly set to `CLOSED`

Here is a simple driver, using pin 17 (BCM)/ 11(BOARD) for the switch, using the new QR stuff of course ;)

```
from pyhap.accessories.GarageDoor import GarageDoor
from pyhap.accessory_driver import AccessoryDriver
import signal


garage = GarageDoor(2, 17, 'GarageDoorTest', pincode=b'920-54-727')
print(garage.display_name)
print(garage.pincode.decode())
garage.print_qr()
driver = AccessoryDriver(garage, port=51826)
signal.signal(signal.SIGINT, driver.signal_handler)
signal.signal(signal.SIGTERM, driver.signal_handler)
driver.start()
```

Please see PR

schinckel:
Oh, oh, oh, I made a garage door opener. You need to do some fancy stuff with target state and current state.

jslay88:
Teach me, obi wan

gerswin:
[https://github.com/gerswin/HAP-python/blob/master/pyhap/accessories/mqtt/GarageDoor.py](url)
check this, maybe can help you

ikalchev:
Closing this as it was already discussed. Keep the conversation open if needed.

PR #36, opened 2018-02-17
Initial RPi Garage Door Commit

jslay88:
Some issues, please see issue #35


pyhap/accessories/GarageDoor.py
```diff
@@ -0,0 +1,184 @@
+import time
+import threading
+import pyhap.loader as loader
+import RPi.GPIO as GPIO
```
schinckel: If you use `gpiozero`, then it's possible to "swap out" the pin factory - this makes testing from another machine (where you have `pigpio` remote pins).

jslay88: Will look into it tomorrow or next. Need to rest  :)

schinckel: https://bitbucket.org/schinckel/garagedoor/src/e3268a75ca576f7dc65532adf1f6d51a8bc149fc/garage_door/accessory.py?at=default&fileviewer=file-view-default

schinckel: Oh, the other great thing about `gpiozero` is that you can set event handlers on pins.

In other words: "when the top trigger opens, this means the door is now closing", and so on.

jslay88: Yeah, thats basically what the callbacks are doing using the GPIO library. I only have one switch on my door, so I infer a lot here, on state and transition. lol.

jslay88: My excuse to learn a lot of different frameworks.

https://gitlab.com/jslay/garage_hub

schinckel: I'm happy to have one that works (a) with no switches: time delay based only, (b) a single switch, and (c) switches at both the top and bottom.

jslay88: yeah, thats one thing i was thinking about, how best to make this Accessory dynamic. 

I just started getting confused about what was happening with my issues listed above, that I figured I'd go ahead and post what I had so far.

ikalchev:
Is this more as a need-info pull request?

jslay88:
Yes, would possibly like to merge with master once refined.

jslay88:
Not sure. Feel free to take what he has if you’d like.

I mainly needed a software debounce method as all of the GPIO libraries
that do support debounce for Pi, have the same issue with latent functions
as the callback.

Then was having weird issues and needed some help since I don’t have
documentation on how these are expected to work.

jslay88:
I have no problem adding onto what he has done to get what I need out of
it. I think we were also beginning to discuss what we want in the
GarageDoor Accessory. So we are likely to build off of whatever gets
merged. We just want to provide you with what we are working on, just funny
we happen to literally did the 2 same things with no background each other.
Lol

ikalchev:
Haha okay. Well, you both are driving this, so it would be perfect if you can say how you prefer to do the merges :)

I see that #42 is at a very finished state with docs and stuff, but it is not for GPIO specifically. Perhaps we can again split the work in two changes, i.e. 42 and your DoorSwitch here (Correct me if I am misunderstanding the commits here, I haven't had the chance to get into it more deeply).

As I said, I have no preference, so if you two have a plan on how to merge these or need help, just let me know :)

jslay88:
Yeah for sure his is in a more complete state. I’d say take his for the
ground work and I can work with him to add in what I need out of it.

ikalchev:
Sounds good. There is just one question from my side at the review, so its pretty much there.

PR #39, opened 2018-02-17
Allow easy way to store state files in user directories.

schinckel:
`..., persist_state='~/.foo_bar.state')`


pyhap/accessory_driver.py
```diff
@@ -145,7 +145,7 @@ def __init__(self, accessory, port, address=None, persist_file="accessory.state"
         self.http_server_thread = None
         self.advertiser = Zeroconf()
         self.port = port
-        self.persist_file = persist_file
+        self.persist_file = os.path.expanduser(persist_file)
```
ikalchev: I don't mind, but add the expansion behaviour to the docs of `persist_file`

PR #41, opened 2018-02-17
Allow classes to easily override AccessoryInformation data.

### LINKED ISSUE #37: AccessoryService information.
I'm not exactly sure how to override the `AccessoryService` fields: I've tried a bit in my `GarageDoor._set_services()`, but it doesn't seem to work.

I wonder if there should be a way to have this information either pulled from the class, or some other mechanism.

schinckel:
Resolves #37.

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -183,6 +183,9 @@ class Accessory(object):
             self._pincode = util.generate_pincode()
         return self._pincode
 
+    def get_accessory_information(self):
+        return {}
+
     def _set_services(self):
         """Sets the services for this accessory.
 
@@ -193,15 +196,29 @@ class Accessory(object):
             always call the base implementation first, as it reserves IID of
             1 for the Accessory Information service (HAP requirement).
         """
+        info_data = {
+            'Name': self.display_name,
+            'Manufacturer': 'Default Manufacturer',
+            'Model': 'Default Model',
+            'SerialNumber': 'Default SerialNumber'
+        }
+        info_data.update(**self.get_accessory_information())
+
         info_service = get_serv_loader().get("AccessoryInformation")
-        info_service.get_characteristic("Name")\
-                    .set_value(self.display_name, False)
-        info_service.get_characteristic("Manufacturer")\
-                    .set_value("Default-Manufacturer", False)
-        info_service.get_characteristic("Model")\
-                    .set_value("Default-Model", False)
-        info_service.get_characteristic("SerialNumber")\
-                    .set_value("Default-SerialNumber", False)
+        for name, value in info_data.items():
+            try:
+                info_service.get_characteristic(name).set_value(value, False)
+            except KeyError:
+                pass
+
+        # info_service.get_characteristic("Name")\
+        #             .set_value(self.display_name, False)
+        # info_service.get_characteristic("Manufacturer")\
+        #             .set_value("Default-Manufacturer", False)
+        # info_service.get_characteristic("Model")\
+        #             .set_value("Default-Model", False)
+        # info_service.get_characteristic("SerialNumber")\
+        #             .set_value("Default-SerialNumber", False)
         # FIXME: Need to ensure AccessoryInformation is with IID 1.
         self.add_service(info_service)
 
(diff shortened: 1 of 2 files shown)


pyhap/accessory.py
```diff
@@ -193,15 +196,29 @@ def _set_services(self):
             always call the base implementation first, as it reserves IID of
             1 for the Accessory Information service (HAP requirement).
         """
+        info_data = {
```
ikalchev: Shouldn't this be the default result of `get_accessory_information`? Otherwise, if you call that method, you will not get the actual accessory information in the default (not override) case.

pyhap/accessory.py
```diff
@@ -193,15 +196,29 @@ def _set_services(self):
...
-                    .set_value(self.display_name, False)
-        info_service.get_characteristic("Manufacturer")\
-                    .set_value("Default-Manufacturer", False)
-        info_service.get_characteristic("Model")\
-                    .set_value("Default-Model", False)
-        info_service.get_characteristic("SerialNumber")\
-                    .set_value("Default-SerialNumber", False)
+        for name, value in info_data.items():
+            try:
+                info_service.get_characteristic(name).set_value(value, False)
```
ikalchev: As of yesterday, you can now safely remove the False in `set_value`.

pyhap/accessory.py
```diff
@@ -193,15 +196,29 @@ def _set_services(self):
...
+            except KeyError:
+                pass
```
ikalchev: Do we really want this behaviour? Imagine I misspell `SerialNUmber` in `get_accessory_information` - I will get no indication that I am referring to a characteristic that does not exist.

I would remove the `try` and fail in case of a missing characteristic. (I am assuming that `get_acc_info` is specifically for configuring the AccessoryInformation service)

pyhap/accessory.py
```diff
@@ -193,15 +196,29 @@ def _set_services(self):
...
+
+        # info_service.get_characteristic("Name")\
```
ikalchev: Yeah just shoot this.

pyhap/accessory.py
```diff
@@ -183,6 +183,9 @@ def pincode(self):
             self._pincode = util.generate_pincode()
         return self._pincode
 
+    def get_accessory_information(self):
```
ikalchev: Could you please write the docs here? E.g. make it explicit that this method is specifically aimed at configuring the AccInfo Service (if that's your idea).

pyhap/accessory.py
schinckel: Yeah, I wasn't sure about this either. Perhaps just having no exception handling will be easiest.

pyhap/accessory.py
schinckel: I was trying to get away from having to call `super()` in that method. Maybe it should be something about custom accessory information.

schinckel: I also wonder if we could get rid of the `super()` requirement from `_set_services()`, or look at some other way of defining an extra service.

Perhaps an `@add_service()` decorator, which sets up a service and also acts as a property - so we could have nice definitions. Just thought of this now, will write up an example later.

ikalchev: Why are you trying to get away from the `super`?

schinckel: Mainly because it's easy to accidentally forget to call `super()` (or not call it until after other stuff has happened), whereas this code _must_ run (ie, AccessoryInformation _must_ be IID=1).

I think it would be better to have a hook that you can just override, or some other way of registering services. I'll think more about that.

schinckel: I also think that a simpler function that returns just the things that we want to override would make implementing that method (in the subclass) simpler.

pyhap/accessory.py
```diff
@@ -183,6 +183,17 @@ def pincode(self):
...
```
ikalchev: Okay I get your idea. Can we then call this `create_accessory_information` or something similar? My issues is that `get` intuitively will mean "get me the current accessory information", whereas it should be "create my custom accessory information". Does it make sense?

pyhap/accessory.py
ikalchev: I see your point, I will try to address this.

The IID=1 requirement is a big hack as is currently implemented. I will surely kick this out from here and make something that is harder to screw up.

cdce8p:
This PR should no longer be necessary since #102 should cover it.

PR #42, opened 2018-02-17
Garage door

schinckel:
GarageDoor plugin using `gpiozero` (which is really cool, by the way).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,4 @@ __pycache__
 # HAP-python-generated files
 accessory.pickle
 accessory.state
-
+*.state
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -18,4 +18,4 @@ __pycache__
 # HAP-python-generated files
 accessory.pickle
 accessory.state
-
+*.state
```
ikalchev: If you are adding this, remove `accessory.state` from above.

tox.ini
```diff
@@ -2,9 +2,19 @@
 envlist = py35,py36
 
 [testenv]
+basepython = python3.6
```
ikalchev: Could you please elaborate on what happens if there is no python3.6? I am not tox-fluent enough.

ikalchev:
What is the state of this? Do you want me to resolve the conflicts and merge?

PR #55, opened 2018-02-21
RTD Import Fix

jslay88:
Under the project panel, Admin -> Advanced Settings

- We need to have `Install your project inside a virtualenv using setup.py install` checked.
- `Python interpreter` needs to be set to `CPython3.x`

See here for a successful RTD build.
http://hap-python-test.readthedocs.io/en/rtdimportfix/


jslay88:
Under the project panel, Admin -> Advanced Settings

- We need to have `Install your project inside a virtualenv using setup.py install` checked.
- `Python interpreter` needs to be set to `CPython3.x`

PR #56, opened 2018-02-23
Added object representations

cdce8p:
For `Accessory`, `Service` and `Characteristic`  
Added python venv suff to `.gitignore`.

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -165,6 +165,12 @@ class Accessory(object):
 
         self._set_services()
 
+    def __repr__(self):
+        """Return the representation of the accessory."""
+        services = [s.display_name for s in self.services]
+        return "<accessory display_name='{}' services={}>" \
+            .format(self.display_name, services)
+
     def __getstate__(self):
         state = self.__dict__.copy()
         state["broker"] = None
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -339,9 +339,9 @@ class Accessory(object):
         print(self.qr_code.terminal(), flush=True)
 
     def get_characteristic(self, aid, iid):
-        """Get's the characteristic for the given IID.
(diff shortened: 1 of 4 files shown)


pyhap/accessory.py
```diff
@@ -165,6 +165,12 @@ def __init__(self, display_name, aid=None, mac=None, pincode=None,
 
         self._set_services()
 
+    def __repr__(self):
+        """Return the representation of the accessory."""
+        services = [s.display_name for s in self.services]
+        return "<accessory display_name='{}' services={}>" \
```
ikalchev: Consistency using ticks around properties. Also I would write a capital A (for Accessories)

pyhap/characteristic.py
```diff
@@ -114,6 +114,11 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
         self.setter_callback = None
         self.hap_template = self._create_hap_template()
 
+    def __repr__(self):
+        """Return the representation of the characteristic."""
+        return "<characteristic display_name='{}' value={} properties={}>" \
+            .format(self.display_name, self.display_name, self.properties)
```
ikalchev: The second item should be `self.value`. Again the ticks (I get that you want to handle spaces in the display name, but maybe they could also prop in the value or properties)

ikalchev:
Okay, let's leave the ticks as it is. For the names - I would either keep them the same or use a capital. I think it is more intuitive to expect to see the class name of some sort in the repr.

cdce8p:
Wouldn't matter to me. What do you like more?

ikalchev:
Honestly, capitals (`Accessory`, etc.). But I don't mind the way they are now, so whatever.

cdce8p:
I personally would use the representation for debugging, so it wouldn't matter either :smile: 

I changed `self.display_state -> self.value`.

PR #57, opened 2018-02-25
Char: Set default for 'valid_value' to 0

cdce8p:
A `valid_value` set to 0 makes testing easier. Before the `next` function could return any of the values, since (at least for py35) `dict.values` doesn't necessarily returns a sorted list.

The value `0` is set for all characteristics that have the attribute `validValues` so it shouldn't be an issue.

### DIFF
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -34,6 +34,7 @@ class HAP_FORMAT:
         UINT64: 0,
         DATA: "",
         TLV8: "",
+        'VALID_VALUES': 0,
     }
 
 
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -34,7 +34,6 @@ class HAP_FORMAT:
         UINT64: 0,
         DATA: "",
         TLV8: "",
-        'VALID_VALUES': 0,
     }
 
 
@@ -107,7 +106,7 @@ class Characteristic(object):
             self.value = value
         else:
             if self.properties.get('ValidValues'):
-                self.value = next(iter(self.properties["ValidValues"].values()))
+                self.value = min(self.properties['ValidValues'].values())
             else:
                 self.value = HAP_FORMAT.DEFAULT[properties["Format"]]
         self.broker = broker


pyhap/characteristic.py
```diff
@@ -105,7 +106,7 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
         self.has_valid_values = "ValidValues" in self.properties
         if value is None:
             if self.has_valid_values:
-                self.value = next(iter(self.properties["ValidValues"].values()))
+                self.value = HAP_FORMAT.DEFAULT['VALID_VALUES']
```
ikalchev: I get the point of this change, however this will break all Characteristics which don't have 0 as a valid value. It's true there are no such chars now, but this means we need to monitor for such changes (in pyhap/resources) in the future.

Let me clarify: the idea of this code is to just put some valid value, because if iOS makes `get_characteristics` and receives an invalid value, it will unpair with the Accessory right away.

Thoughts?

cdce8p: I get what you mean, but (as far as I know) there are also no checks in `CharLoader` if the entry `validValues` contained at least one item.
Setting it to `0` by default even prevents an error in these cases, since there are also no checks if the `set_value` is in `validValues`.

ikalchev: Actually there is - on line 154.
```python
if (self.has_valid_values and value not in self.properties["ValidValues"].values()):
    raise ValueError
```
I don't mind doing that but it has the chance of breaking someone when new Chars are released - and what we gain is easier testing for the default value (correct?). On another topic, If we decide to do that, I would use a value from `HAP_FORMAT.DEFAULT` because it will be from the same value type.

ikalchev: My reasoning was that you can use set value in a broader context, not just for HAP. Imagine you want to log your values in a DB - you would probably want all values there, no truncating applied. You can just plug into the setter and do that there. Does it make sense or is it an overkill?

ikalchev: Writing test will be harder but using the framework will be easier as default valid values are set for you. On the other hand, as all valid values are ints and can be 0, we may use that shortcut. I am trying to look at it from all sides before we can decide.

cdce8p: Regarding the DB comment: At least I personally wouldn't do it that way. Calling `set_value` in the first place requires to have a value. Why not log it there, instead of calling `get_value` again. Hope that makes sense.

cdce8p: What do you think of `self.value = sorted(self.properties["ValidValues"].values())[0]` instead?

ikalchev: What about `self.value = min(self.properties["ValidValues"].values())`?

cdce8p: Sounds good. Do we want to add `, 0` to the min function?

ikalchev: You mean for the default?

cdce8p:
Yea: `self.value = min(self.properties["ValidValues"].values(), 0)`

pyhap/characteristic.py
cdce8p: You're right. I missed that.

cdce8p:
I changed the statement use `min` and rebased the branch against dev to solve merge conflicts. Should be good now.

ikalchev:
Thanks for the many contributions and suggestions for improvements! Keep them coming.

ikalchev:
Ok I am on discord as well now

PR #62, opened 2018-03-22
Acc Driver: Should_notify

cdce8p:
I recently came across an issue on the Home Assistant Github page: https://github.com/home-assistant/home-assistant/issues/13345

A user has two iPads and if he executes a command on one the state doesn't change on the second one. I believe that can be fixed with setting `should_notify=True` in the `set_characteristics` method of `accessory_driver`. However I don't know if there are any reasons why you choose to set it in the first place.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -440,7 +440,7 @@ class AccessoryDriver(object):
             }
             if "value" in cq:
                 # TODO: status needs to be based on success of set_value
-                char.set_value(cq["value"], should_notify=False)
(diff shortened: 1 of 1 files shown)


ikalchev:
If I recall correctly this made iOS unhappy (unsupported acc or unpaired), but I need to double check this. Have you tried if this change works as expected with, say, (mock) temperature sensor or a bulb?

cdce8p:
I didn't noticed any issues with the accessories I've tested including the `(mock) temperature sensor`.

PR #63, opened 2018-03-22
Char Set value: setter_callback

cdce8p:
I believe that the callback method shouldn't be called if a char is set to the same value it has been before. For some this might be a breaking change.

What do you think?

### DIFF
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -153,9 +153,10 @@ class Characteristic(object):
         if (self.has_valid_values
                 and value not in self.properties["ValidValues"].values()):
             raise ValueError
-        self.value = value
-        if self.setter_callback is not None and should_callback:
+        if self.setter_callback is not None and should_callback and \
+                value != self.value:
             self.setter_callback(value)
+        self.value = value
         if should_notify and self.broker is not None:
             self.notify()
 


ikalchev:
Hey, it's been a while!

There are some chars for which setting the same value twice makes sense, but these are more of an exception. In that case, the optimisation you are proposing makes sense, however, I would like for users to be able to disable this check. Also, note that in the current design, this same functionality can be achieved in the `setter_callback` - implies that the `setter_callback` is stateful. The last bit can be removed if callbacks also had access to the `self` of course. What do you think?

Can you give me more context on the reason for this request, i.e. is it performance that bothers you or is it usability, so that we can work better towards a solution?

Best,
Ivan

cdce8p:
It was more meant as a general suggestion and to prevent potential errors that might occur if `setter_callback=false` isn't set.

In a broader context: I'm still looking for a good solution to handle the case when both (HomeKit and the HAP server) can update a char value and and the setter method calls an action on the server which in turn would call an action on HomeKit, so on. My current solution using flag variables for each char feels overly complicated. An example for this would be: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/homekit/type_lights.py and how a brightness change is handled in the [setter_callback](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/homekit/type_lights.py#L80-L85) and the [update](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/homekit/type_lights.py#L125-L133) method.

cdce8p:
See #73 for follow up

PR #65, opened 2018-03-23
Added requirements file

cdce8p:
Makes it easier for users to install all needed dependencies.


ikalchev:
I am missing knowledge here - How is this different than running `pip3 install HAP-python` and the list of dependencies in the setup.py file?

cdce8p:
You're right, missed that. `pip3 install -r requirements_all.txt` is only needed for development, when copying the Github repo instead of `pip3 install HAP-python`. I'll fix that.

PR #66, opened 2018-03-23
Char: override_properties

cdce8p:
Related discussion: #26

* Removed `_create_hap_template` an merged this into `to_HAP`
* New method `override_properties` for chars
* Removed `self.has_valid_values` and replaced it with runtime checks to prevent update errors

### DIFF
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -102,14 +102,13 @@ class Characteristic(object):
         self.display_name = display_name
         self.type_id = type_id
         self.properties = properties
-        self.has_valid_values = "ValidValues" in self.properties
-        if value is None:
-            if self.has_valid_values:
+        if value:
+            self.value = value
+        else:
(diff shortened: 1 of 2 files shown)


pyhap/characteristic.py
```diff
@@ -166,6 +153,23 @@ def get_value(self):
...
+        :type properties: dict
+
+        :param valid_values: Dictionary with values to override the existing
+            valid_values. Valid values will be set to new dictionary.
+        :type valid_values: dict
+        """
+        if properties:
+            self.properties.update(properties)
+
+        if valid_values:
```
ikalchev: I would add:
```
if valid_values:
    if not "ValidValues" in self.properties:
        raise ValueError(...)
    else:
        ...
```

ikalchev: Actually forget the above - I will check if you can set ValidValues on chars that don't have ValidValues defined and will get back to you or will submit a change myself.

PR #67, opened 2018-03-25
Remove opt_chars and get_value

cdce8p:
A bit cleanup.
Removed:
* opt_characteristics (service)
* add_opt_characteristics (service)
* get_value (chars)

## Breaking changes:
* Removed `add_opt_characteristic`. There is no further distinction between optional and require chars during execution. All chars that have been added as `optional` should now simply use `add_characteristic`. (The loader hasn't changed.)
* The method `get_characteristic` now raises a `ValueError` if the characteristic isn't found
* The already deprecated method `get_value` was removed. Use `char.value` instead.

### DIFF
--- a/pyhap/accessories/FakeFan.py
+++ b/pyhap/accessories/FakeFan.py
@@ -28,12 +28,12 @@ class FakeFan(Accessory):
 
         # Add the optional RotationSpeed characteristic to the Fan
         rotation_speed_char = loader.get_char_loader().get("RotationSpeed")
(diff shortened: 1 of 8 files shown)


ikalchev:
This is just perfect.

I had doubts about removing the `get_value`, but we can bring it back if a use case shows up for it.

cdce8p:
I don't think that it will. `char.value` should cover it and is more pythonic :)

PR #72, opened 2018-03-30
Removed 'update_advertisment' from 'pair'

cdce8p:
Temporary workaround for https://github.com/home-assistant/home-assistant/issues/12868

@ikalchev What do you think of squash merging PR's (https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits)? IMO it would help keeping the commit history clean.


ikalchev:
Yes, I will squashing now.

ikalchev:
I would prefer if you comment out the method call and put a comment pointing to the PR. I would really like to chase this down some day.

pyhap/accessory_driver.py
```diff
@@ -510,8 +528,10 @@ def stop(self):
         self.accessory_thread.join()
 
         logger.debug("Stopping mDNS advertising")
-        self.advertiser.unregister_service(self.mdns_service_info)
```
cdce8p: `unregister_service` not necessary, since already included in `advertiser.close()`

pyhap/accessory_driver.py
```diff
@@ -254,20 +259,27 @@ def config_changed(self):
...
+        except:  # KeyError (not propagated by zeroconf)
+            logger.error(WARNING_MSG_UNREGISTER_SERVICE)
         self.mdns_service_info = AccessoryMDNSServiceInfo(self.accessory,
                                                           self.address,
                                                           self.port)
         time.sleep(0.1)  # Doing it right away can cause crashes.
-        self.advertiser.register_service(self.mdns_service_info)
+        try:
+            self.advertiser.register_service(self.mdns_service_info)
+        except:  # NonUniqueNameException (not propagated by zeroconf)
```
ikalchev: why don't we except only for NonUniqueName?

pyhap/accessory_driver.py
```diff
@@ -486,7 +498,11 @@ def start(self):
         self.mdns_service_info = AccessoryMDNSServiceInfo(self.accessory,
                                                           self.address,
                                                           self.port)
-        self.advertiser.register_service(self.mdns_service_info)
+        try:
+            self.advertiser.register_service(self.mdns_service_info)
+        except:
+            logger.error('Could not register HomeKit service. Restart')
+            self.stop()
```
ikalchev: add a return after the stop otherwise the QR code will get printed

pyhap/accessory_driver.py
```diff
@@ -510,8 +529,10 @@ def stop(self):
         self.accessory_thread.join()
 
         logger.debug("Stopping mDNS advertising")
-        self.advertiser.unregister_service(self.mdns_service_info)
-        self.advertiser.close()
+        try:
+            self.advertiser.close()
+        except:
+            pass
```
ikalchev: Can we log the exception here? it will help in debugging

pyhap/accessory_driver.py
```diff
@@ -502,6 +518,9 @@ def stop(self):
         """
         # TODO: This should happen in a different order - mDNS, server, accessory. Need
         # to ensure that sending with a closed server will not crash the app.
+        if self.run_sentinel.is_set():
```
ikalchev: could you elaborate on this?

pyhap/accessory_driver.py
cdce8p: I only got: `Unknown error, possibly begin NonUniqueName`. When I only tried to catch the `NonUniqueName`, it failed.

pyhap/accessory_driver.py
cdce8p: Would a debug statement be enough, since it isn't relevant for any implementation?

ikalchev: Sounds good

pyhap/accessory_driver.py
cdce8p: `stop()` doesn't need to be executed again if the HAP server is already stopped, because register service failed.

pyhap/accessory_driver.py
ikalchev: How do you test? The NonUnique should propagate from here https://github.com/jstasiak/python-zeroconf/blob/master/zeroconf.py#L1875

pyhap/accessory_driver.py
ikalchev: Ah I see, yes

pyhap/accessory_driver.py
cdce8p: I modified the `main.py` to:

```py
def run(self):
    self.temp_char.set_value(20)
    while not self.run_sentinel.wait(0.5):
        print('==> Config Changed')
        self.config_changed()
```

That seems to reliably produce these errors and the second one isn't caught. Although that doesn't happen with only `except:` either. I change it.

cdce8p: Interestingly if I add `self.run_sentinel.wait(1)` belove `set_value` I don't get these errors.

ikalchev:
Looks good, let's see @iditude's experience.

ikalchev:
Hey, I sort of got lost in the other issue. How is this going, is this change solving the problem?

cdce8p:
The last update I know of was that it was working with the modified version for `1.1.9`.
I'll update this to `2.0.0` and ask how it's working.

PR #73, opened 2018-03-31
Characteristic improvements

cdce8p:
I started working on `characteristic.py` and got some first results.

Thats what I have planed:
 - [x] Merge `set_value` and `get_hap_value`
 - [x] Set iid_manager as instance property, not parameter for `to_HAP`

This should lead to:
 - Add the classmethod `from_dict` --> later
 - Change loader to use `from_dict` --> later

### Breaking changes
- `set_value` doesn't take `should_callback` anymore

### DIFF
--- a/pyhap/accessories/SDS011.py
+++ b/pyhap/accessories/SDS011.py
@@ -73,20 +73,20 @@ class SDS011(Accessory):
         # PM2.5
         air_quality_pm25 = loader.get_serv_loader().get("AirQualitySensor")
         pm25_size = char_loader.get("AirParticulateSize")
-        pm25_size.set_value(0, should_notify=False)
+        pm25_size.value = 0
         self.pm25_density = char_loader.get("AirParticulateDensity")
         pm25_name = char_loader.get("Name")
-        pm25_name.set_value("PM2.5", should_notify=False)
+        pm25_name.value = "PM2.5"
         self.pm25_quality = air_quality_pm25.get_characteristic("AirQuality")
         air_quality_pm25.add_characteristic(pm25_name, pm25_size, self.pm25_density)
 
         # PM10
         air_quality_pm10 = loader.get_serv_loader().get("AirQualitySensor")
         pm10_size = char_loader.get("AirParticulateSize")
-        pm10_size.set_value(1, should_notify=False)
+        pm10_size.value = 1
         self.pm10_density = char_loader.get("AirParticulateDensity")
         pm10_name = char_loader.get("Name")
-        pm10_name.set_value("PM10", should_notify=False)
+        pm10_name.value = "PM10"
         self.pm10_quality = air_quality_pm10.get_characteristic("AirQuality")
         air_quality_pm10.add_characteristic(pm10_name, pm10_size, self.pm10_density)
 
--- a/pyhap/accessories/SDS011.py
+++ b/pyhap/accessories/SDS011.py
@@ -73,20 +73,20 @@ class SDS011(Accessory):
         # PM2.5
         air_quality_pm25 = loader.get_serv_loader().get("AirQualitySensor")
         pm25_size = char_loader.get("AirParticulateSize")
-        pm25_size.value = 0
+        pm25_size.set_value(0, should_notify=False)
         self.pm25_density = char_loader.get("AirParticulateDensity")
         pm25_name = char_loader.get("Name")
-        pm25_name.value = "PM2.5"
+        pm25_name.set_value("PM2.5", should_notify=False)
         self.pm25_quality = air_quality_pm25.get_characteristic("AirQuality")
         air_quality_pm25.add_characteristic(pm25_name, pm25_size, self.pm25_density)
 
         # PM10
         air_quality_pm10 = loader.get_serv_loader().get("AirQualitySensor")
         pm10_size = char_loader.get("AirParticulateSize")
-        pm10_size.value = 1
+        pm10_size.set_value(1, should_notify=False)
         self.pm10_density = char_loader.get("AirParticulateDensity")
         pm10_name = char_loader.get("Name")
-        pm10_name.value = "PM10"
+        pm10_name.set_value("PM10", should_notify=False)
         self.pm10_quality = air_quality_pm10.get_characteristic("AirQuality")
         air_quality_pm10.add_characteristic(pm10_name, pm10_size, self.pm10_density)
 
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -109,20 +109,20 @@ class Characteristic:
         """Perform validation and conversion to valid value"""
         if self.properties.get('ValidValues'):
             if value not in self.properties['ValidValues'].values():
-                logger.error('%s: value={} is an invalid value.',
-                             self.display_name, value)
-                raise ValueError('{}: value={} is an invalid value.'
-                                 .format(self.display_name, value))
+                error_msg = '{}: value={} is an invalid value.' \
+                            .format(self.display_name, value)
+                logger.error(error_msg)
+                raise ValueError(error_msg)
         elif self.properties['Format'] == HAP_FORMAT.STRING:
             value = str(value)[:256]
         elif self.properties['Format'] == HAP_FORMAT.BOOL:
             value = bool(value)
         elif self.properties['Format'] in HAP_FORMAT.NUMERIC:
             if not isinstance(value, (int, float)):
-                logger.error('%s: value=%s is not a numeric value.',
-                             self.display_name, value)
-                raise ValueError('{}: value={} is not a numeric value.'
-                                 .format(self.display_name, value))
+                error_msg = '{}: value={} is not a numeric value.' \
+                            .format(self.display_name, value)
+                logger.error(error_msg)
+                raise ValueError(error_msg)
             value = min(self.properties.get('maxValue', value), value)
             value = max(self.properties.get('minValue', value), value)
         return value
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -100,6 +100,9 @@ class Characteristic:
     def set_value(self, value, should_notify=True):
         """Set the given raw value. It is checked if it is a valid value.
 
+        If not set_value will be aborted and an error message will be
+        displayed.
+
         :param value: The value to assign as this Characteristic's value.
         :type value: Depends on properties["Format"]
 
@@ -109,6 +112,8 @@ class Characteristic:
         """
         logger.debug('%s: Set value to %s', self.display_name, value)
         value = self.to_valid_value(value)
+        if value is None:
+            return
         self.value = value
         if should_notify and self.broker:
             self.notify()
@@ -125,12 +130,18 @@ class Characteristic:
         """Perform validation and conversion to valid value"""
         if self.properties.get('ValidValues'):
             if value not in self.properties['ValidValues'].values():
-                raise ValueError('%s is not a valid value'.format(value))
+                logger.error('%s: value=%s is an invalid value.',
+                             self.display_name, value)
+                return
         elif self.properties['Format'] == HAP_FORMAT.STRING:
             value = str(value)[:256]
         elif self.properties['Format'] == HAP_FORMAT.BOOL:
             value = bool(value)
         elif self.properties['Format'] in HAP_FORMAT.NUMERIC:
+            if not isinstance(value, (int, float)):
+                logger.error('%s: value=%s is not a numeric value.',
+                             self.display_name, value)
+                return
             value = min(self.properties.get('maxValue', value), value)
             value = max(self.properties.get('minValue', value), value)
         return value
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -6,7 +6,7 @@ a temperature measuring or a device status.
 """
 import logging
 
-_LOGGER = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
 
 class HAP_FORMAT:
@@ -107,7 +107,7 @@ class Characteristic:
             subscribed clients. Notify will be performed if the broker is set.
         :type should_notify: bool
         """
-        _LOGGER.debug('%s: Set value to %s', self.display_name, value)
+        logger.debug('%s: Set value to %s', self.display_name, value)
         value = self.to_valid_value(value)
         self.value = value
         if should_notify and self.broker:
@@ -165,7 +165,7 @@ class Characteristic:
 
         Change self.value to value and call callback.
         """
-        _LOGGER.debug('%s: Client update value to %s',
+        logger.debug('%s: Client update value to %s',
                       self.display_name, value)
         self.value = value
         self.notify()
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -128,6 +128,8 @@ class Characteristic:
         """Perform validation and conversion to valid value"""
         if self.properties.get('ValidValues'):
             if value not in self.properties['ValidValues'].values():
+                logger.error('%s: value={} is an invalid value.',
+                             self.display_name, value)
                 raise ValueError('{}: value={} is an invalid value.'
                                  .format(self.display_name, value))
         elif self.properties['Format'] == HAP_FORMAT.STRING:
@@ -136,6 +138,8 @@ class Characteristic:
             value = bool(value)
         elif self.properties['Format'] in HAP_FORMAT.NUMERIC:
             if not isinstance(value, (int, float)):
+                logger.error('%s: value=%s is not a numeric value.',
+                             self.display_name, value)
                 raise ValueError('{}: value={} is not a numeric value.'
                                  .format(self.display_name, value))
             value = min(self.properties.get('maxValue', value), value)
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -4,6 +4,11 @@ All things for a HAP characteristic.
 A Characteristic is the smallest unit of the smart home, e.g.
 a temperature measuring or a device status.
 """
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+
 class HAP_FORMAT:
     BOOL = 'bool'
     INT = 'int'
@@ -55,13 +60,6 @@ class CharacteristicError(Exception):
     pass
 
 
-_HAP_NUMERIC_FIELDS = ('maxValue', 'minValue', 'minStep', 'unit')
-"""Fields that should be included in the HAP representation of the characteristic.
-
-That is, if they are present in the specification of a numeric-value characteristic.
-"""
-
-
 class Characteristic:
     """Represents a HAP characteristic, the smallest unit of the smart home.
 
@@ -109,6 +107,7 @@ class Characteristic:
             subscribed clients. Notify will be performed if the broker is set.
         :type should_notify: bool
         """
+        _LOGGER.debug('%s: Set value to %s', self.display_name, value)
         value = self.to_valid_value(value)
         self.value = value
         if should_notify and self.broker:
@@ -166,6 +165,8 @@ class Characteristic:
 
         Change self.value to value and call callback.
         """
+        _LOGGER.debug('%s: Client update value to %s',
+                      self.display_name, value)
         self.value = value
         self.notify()
         if self.setter_callback:
@@ -189,7 +190,8 @@ class Characteristic:
 
         if self.properties['Format'] in HAP_FORMAT.NUMERIC:
             hap_rep.update({k: self.properties[k] for k in
-                              self.properties.keys() & _HAP_NUMERIC_FIELDS})
+                            self.properties.keys() & 
+                            ('maxValue', 'minValue', 'minStep', 'unit')})
         elif self.properties['Format'] == HAP_FORMAT.STRING:
             if len(self.value) > 64:
                 hap_rep['maxLen'] = min(len(self.value), 256)
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -56,21 +56,14 @@ class CharacteristicError(Exception):
     pass
 
 
-class NotConfiguredError(Exception):
-    """Raised when an operation is attempted on a characteristic that has not been
-    fully configured.
-    """
-    pass
-
-
-_HAP_NUMERIC_FIELDS = {"maxValue", "minValue", "minStep", "unit"}
+_HAP_NUMERIC_FIELDS = ('maxValue', 'minValue', 'minStep', 'unit')
 """Fields that should be included in the HAP representation of the characteristic.
 
 That is, if they are present in the specification of a numeric-value characteristic.
 """
 
 
-class Characteristic(object):
+class Characteristic:
     """Represents a HAP characteristic, the smallest unit of the smart home.
 
     A HAP characteristic is some measurement or state, like battery status or
@@ -78,6 +71,9 @@ class Characteristic(object):
     Each characteristic has a unique type UUID and a set of properties,
     like format, min and max values, valid values and others.
     """
+
+    __slots__ = ['display_name', 'type_id', 'properties', 'broker',
+                 'setter_callback', 'value']
 
     def __init__(self, display_name, type_id, properties, value=None, broker=None):
         """Initialise with the given properties.
@@ -98,34 +94,66 @@ class Characteristic(object):
             - else use `HAP_FORMAT.DEFAULT` for the format of this characteristic.
         :type value: Depends on `properties["Format"]`
         """
-        assert "Format" in properties and "Permissions" in properties
+        if 'Format' not in properties or 'Permissions' not in properties:
+            raise ValueError('Invalid properties')
         self.display_name = display_name
         self.type_id = type_id
         self.properties = properties
-        if value:
-            self.value = value
-        else:
-            if self.properties.get('ValidValues'):
-                self.value = min(self.properties['ValidValues'].values())
-            else:
-                self.value = HAP_FORMAT.DEFAULT[properties["Format"]]
         self.broker = broker
         self.setter_callback = None
+        self.value = value or self._get_default_value()
 
     def __repr__(self):
         """Return the representation of the characteristic."""
-        return "<characteristic display_name='{}' value={} properties={}>" \
+        return '<characteristic display_name={} value={} properties={}>' \
             .format(self.display_name, self.value, self.properties)
 
+    def _get_default_value(self):
+        """Helper method. Return default value for format."""
+        if self.properties.get('ValidValues'):
+            return min(self.properties['ValidValues'].values())
+        else:
+            value = HAP_FORMAT.DEFAULT[self.properties['Format']]
+            return self._validate_value(value)
+
+    def _validate_value(self, value):
+        """Perform value validation depending on format."""
+        if self.properties.get('ValidValues'):
+            if value not in self.properties['ValidValues'].values():
+                raise ValueError
+            else:
+                return value
+        elif self.properties['Format'] == HAP_FORMAT.STRING:
+            return str(value)[:256]
+        elif self.properties['Format'] == HAP_FORMAT.NUMERIC:
+            value = min(self.properties.get('maxValue', value), value)
+            return max(self.properties.get('minValue', value), value)
+        elif self.properties['Format'] == HAP_FORMAT.BOOL:
+            return bool(value)
+        elif self.properties['Format'] == HAP_FORMAT.ARRAY:
+            # TODO: Add validation
+            pass
+        elif self.properties['Format'] == HAP_FORMAT.DICTIONARY:
+            # TODO: Add validation
+            pass
+        elif self.properties['Format'] == HAP_FORMAT.DATA:
+            # TODO: Add validation
+            pass
+        elif self.properties['Format'] == HAP_FORMAT.TLV8:
+            # TODO: Add validation
+            pass
+
+        return value
+
     def set_value(self, value, should_notify=True, should_callback=True):
-        """Set the given raw value. It is checked if it is a valid value.
+        """Set the given raw value.
 
         :param value: The value to assign as this Characteristic's value.
         :type value: Depends on properties["Format"]
 
         :param should_notify: Whether a the change should be sent to subscribed clients.
-            The notification is called _after_ the setter callback. Notify will be
-            performed if and only if the broker is set, i.e. not None.
+            The notification is called _after_ the setter callback.
+            Notify will be performed if broker is set.
         :type should_notify: bool
 
         :param should_callback: Whether to invoke the callback, if such is set. This
@@ -137,13 +165,10 @@ class Characteristic(object):
         :raise ValueError: When the value being assigned is not one of the valid values
             for this Characteristic.
         """
-        if self.properties.get('ValidValues') and \
-                value not in self.properties['ValidValues'].values():
-            raise ValueError
-        self.value = value
-        if self.setter_callback is not None and should_callback:
-            self.setter_callback(value)
-        if should_notify and self.broker is not None:
+        self.value = self._validate_value(value)
+        if self.setter_callback and should_callback:
+            self.setter_callback(self.value)
+        if self.broker and should_notify:
             self.notify()
 
     def override_properties(self, properties=None, valid_values=None):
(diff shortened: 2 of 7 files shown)


pyhap/characteristic.py
```diff
@@ -56,21 +56,14 @@ class CharacteristicError(Exception):
...
-_HAP_NUMERIC_FIELDS = {"maxValue", "minValue", "minStep", "unit"}
+_HAP_NUMERIC_FIELDS = ('maxValue', 'minValue', 'minStep', 'unit')
 """Fields that should be included in the HAP representation of the characteristic.
 
 That is, if they are present in the specification of a numeric-value characteristic.
 """
 
 
-class Characteristic(object):
+class Characteristic:
```
cdce8p: `object` inheritance is no longer necessary

pyhap/characteristic.py
```diff
@@ -137,13 +165,10 @@ def set_value(self, value, should_notify=True, should_callback=True):
...
-        if self.properties.get('ValidValues') and \
-                value not in self.properties['ValidValues'].values():
-            raise ValueError
-        self.value = value
-        if self.setter_callback is not None and should_callback:
-            self.setter_callback(value)
-        if should_notify and self.broker is not None:
+        self.value = self._validate_value(value)
+        if self.setter_callback and should_callback:
+            self.setter_callback(self.value)
```
cdce8p: I think the callback should just be a callback and only be called if the value is changed in HomeKit.

pyhap/characteristic.py
```diff
@@ -98,34 +94,66 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
...
+                return value
+        elif self.properties['Format'] == HAP_FORMAT.STRING:
+            return str(value)[:256]
+        elif self.properties['Format'] == HAP_FORMAT.NUMERIC:
+            value = min(self.properties.get('maxValue', value), value)
+            return max(self.properties.get('minValue', value), value)
+        elif self.properties['Format'] == HAP_FORMAT.BOOL:
+            return bool(value)
+        elif self.properties['Format'] == HAP_FORMAT.ARRAY:
+            # TODO: Add validation
```
cdce8p: I don't know if we want to add validation for those cases as well

pyhap/characteristic.py
```diff
@@ -98,34 +94,66 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
...
 
+    def _get_default_value(self):
+        """Helper method. Return default value for format."""
+        if self.properties.get('ValidValues'):
+            return min(self.properties['ValidValues'].values())
+        else:
+            value = HAP_FORMAT.DEFAULT[self.properties['Format']]
+            return self._validate_value(value)
+
+    def _validate_value(self, value):
```
ikalchev: This method actually transforms the value to a valid one, i.e. it is not validation only. Can we rename to `to_valid_value` or something like this?

pyhap/characteristic.py
```diff
@@ -98,34 +94,66 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
...
+        """Perform value validation depending on format."""
+        if self.properties.get('ValidValues'):
+            if value not in self.properties['ValidValues'].values():
+                raise ValueError
+            else:
+                return value
+        elif self.properties['Format'] == HAP_FORMAT.STRING:
+            return str(value)[:256]
+        elif self.properties['Format'] == HAP_FORMAT.NUMERIC:
+            value = min(self.properties.get('maxValue', value), value)
```
ikalchev: Not every char that is numeric has a max or min value. How about:
min and max -> min
min -> min
max -> max
no min and no max -> (probably handled by ValidValues, but for sake of completeness) 0

pyhap/characteristic.py
ikalchev: I wouldn't, it will become needlessly complex.

ikalchev:
BTW why we want to rename to_HAP to to_dict?

pyhap/characteristic.py
thomaspurchas: @cdce8p could you point me to an example where `setter_callback` is used? It strikes me as odd that `setter_callback` is properly exposed or documented. Also I would be inclined to agree with you with regards to not call it.

Also `set_value` is an interesting name. In JS-HAP, there are two APIs `set_value` and `update_value`. `set_value` does not notify anything, just updates the objects internal state. Where as `update_value` will make sure that all the clients are notified of the change.

thomaspurchas: To expand a little more on the above. There doesn't currently appear to be away for an `Accessory` to indicate a value update success or failure.

@ikalchev do you have any docs that explaining how client value updates are processed and forwarded onto `Accessory`s? I would be interested in extending it to allow the `Accessory` objects to indicate failure, and have that sent back to the requesting client.

ikalchev: If I may: You have a bulb, which eventually the user taps on in iOS. This sends a request to HAP-python, which triggers the On characteristic's `set_value` method with `should_callback` as True. The `setter_callback` allows users of HAP-python to plug in and actually turn the bulb on, whether it is by interacting with GPIO or whatever.

About the docs: I haven't documented it, sorry about that. Will update soon.

In HAP-python, `update_value` is the JS `set_value`, whereas internal state can be modified with `self.value`.

thomaspurchas: @ikalchev ok, if I understand this right.

iOS clients cause `set_value` to be called with `should_callback` as `True`, and `Accessory` objects should call `set_value` with `should_callback=False` and `should_notify=True`?

If that's the case wouldn't providing separate APIs be a little cleaner, rather than using flags?

ikalchev: @thomaspurchas Client updates end up in `AccessoryDriver.set_characteristics`, which is where `char.set_value` is called. Currently, there is no (easy) way to indicate failure of this. I can propose two things:
1. `set_value` returns True or False, which is checked in `set_characteristics`
2. `set_value` throws something , which is catches in `set_characteristics`

In both cases the user should be able to plug into `set_value`. So, again, here are a few options:
- Use the current setter_callback - not perfect because the `self.value` is set before this is called. We can of course change the flow.
- Use some other mechanism

Personally, I think changing the flow in `set_value` as
```
...
success = self.callback(new_value)
if not success:
    return False
# continue as before
```
will be flexible enough, but some input will mean a lot. Thoughts?

ikalchev: > Accessory objects should call set_value with should_callback=False and should_notify=True?

It really depends on your use case to be honest. I had discussion with people for this before, but we couldn't figure a better way for it. For example, if your accessory wants to turn on the light, it can just call `set_value(new_value)` to whatever is necessary (with `should_callback=True, should_notify=True`, both are the default) and then leave it to the `setter_callback` to actually turn the bulb on. On the other hand, if turning the lights CAN ALSO happen by some external action that you detect after it's done, then the above approach breaks down, because you have to have more complicated control logic.

The problem is how to handle both cases above well enough. So far I have found these flags acceptable in the cases I've seen, but then what have I seen? If you think the API is missing stuff or if you have a suggestion, I would be more than happy to discuss it and make changes.

ikalchev: I will try to come up with a better way for this once more, but a use case/workflow to think about would be nice.

thomaspurchas: The approach I would use is have `set_value` replaced with `update_value`, which should be called when an iOS client wants the value to change. Then the callback is responsible for trying to make that happen, and raises and exception if it can't.

For the `Accessory` object, I think that a `@property` setter could be used on the `value` attribute, that triggers the notification process.

I think it's very odd for people to be using this API to have the callback triggered. Presumably anything that wants to manipulate the `Characteristic` will do it via the `Accessory` object. In which case the `Accessory` object would be responsible for calling the callback, or only setting `value` once the callback has completed successfully.

That way there is a clear separation of responsibilities, `Characteristic`s only deal with what needs to be presented to iOS. Marshalling `value` sets to iOS clients as notifications, and marshalling requests from iOS to the `Accessory`. The `Accessory` is then responsible for making sure that the `value` stores in a `Characteristic` matches the real world device, regardless how that value was set.

This would also closely match how JS-HAP works as well, and prevents py-HAP from being tangled into peoples accessory code.

Happy to put together some cases/workflows, and also a PR to make that happen. Thought from yourself and @cdce8p welcomed!

pyhap/characteristic.py
cdce8p: Actually that's what I should have done here, or at least intended to do (if it doesn't work).
`get('maxValue', value)` has `value` as default. E.g. no maxValue:
 - get --> value
 - min(value, value) -> value

pyhap/characteristic.py
cdce8p: I will delete those than

pyhap/characteristic.py
cdce8p: Actually I think this would be quite a good usecase for `@property` and `@property.setter` to do `to_valid_value`, since this should be executed every time we try to set a value.

cdce8p:
> BTW why we want to rename to_HAP to to_dict?

I think that it then might be more obvious what this method does just by looking at the method name. Also it would work quiet nicely with `from_dict` :)

cdce8p:
Unfortunately I won't have time tomorrow. I will resume working on this probably Monday.

pyhap/characteristic.py
thomaspurchas: I know I suggested an `@property.setter` above. But thinking about it some more, I am a little concerned that using a `@property.setter` might result in confusing behaviour.

As a user of the API you would write a value to `.value`, and it could either raise an `exception` or when you read `.value` you wouldn't get the value you expected.

Just wondering if instead `.value` should be stored in `._value`, and then expose a `set_value()` and `update_value()` methods which would do the validation. Also using a `@property` getter to allow reading `.value` but not writing to it.

That way, to a user it's clear that when you call `set_value()`/`update_value()` other clever things are gonna happen (including callbacks and notifications). So should be surprised if the value passed is changed, or a validation exception is raised.

cdce8p: After my comment I though about this again an I would agree in many cases @thomaspurchas:
 - a property.setter might indeed be confusing
 - ._value might be a good idea. We use the value variable only as cache for to_HAP if we need to send this to HK. I don't think the user should interact with value directly. _value should be enough, we don't need to explicitly constrain it.
 - @thomaspurchas I understand were your push for two separate methods is comming from. I personally don't like the way it't currently handled. Since should_notify is set to false when update came from HomeKit client, the callback needs to call set_value, or at least notify, so  that all clients have the updated value. Additionally when you want to call set_value, I always need to pass in callback=false to prevent the callback. That is kind of redundant.

cdce8p:
Pushed an update. Sadly the `git diff` got a bit scrambled.
Regarding the current status:
 - As @thomaspurchas suggested, I split up `set_value` in `set_value` and `update_value`. The later should be called if the change came from HomeKit. A byproduct is that `should_callback` and `should_notify` are no longer necessary.
 - `update_value` returns a boolean to indicate if the update was successful (for #75).
 - I added the class method `from_dict` which I think would be a better solution for the char loader. That could be useful for `loader.py` later
 - I decided against implementing something special for the properties, neither a class, nor namedtuples. It just would have produced a lot of overhang with very little benefits.
 - Before we finally merge this, I would like to push a last PR to reorder the class methods. If I would have done that now, the diff would be unreadable.

Still to do:
 - Decide together if this is the way to go.
 - Adding tests

thomaspurchas:
Re `to_HAP` -> `to_dict`. I'm a little concerned about this. `to_HAP` exists to create a representation of the characteristic that the HAP protocol understands. The fact that it's a `dict` is just an implementation detail, and it shouldn't be relied on.

Changing it to `to_dict` give the impression that you could use that representation for something else, and that it's always going to be a `dict` type. Neither of those things are true, in the future we may decide to introduce a `HAPMessage` type or something similar, which would be returned instead of a `dict`. `to_dict` ties our hands on that, conflating what the function does with what it returns.

pyhap/accessories/SDS011.py
```diff
@@ -73,20 +73,20 @@ def _set_services(self):
         # PM2.5
         air_quality_pm25 = loader.get_serv_loader().get("AirQualitySensor")
         pm25_size = char_loader.get("AirParticulateSize")
-        pm25_size.set_value(0, should_notify=False)
+        pm25_size.value = 0
```
thomaspurchas: I though that accessories should be calling `set_value`, and in the future `value` will be made read only with a `@property` getter.

pyhap/characteristic.py
```diff
@@ -163,70 +150,45 @@ def override_properties(self, properties=None, valid_values=None):
...
-                                     "Consider adding the characteristic to a "
-                                     "Service and then to an Accessory.")
-
-        data = {
-            "type_id": self.type_id,
-            "value": self.get_hap_value(),
+        """Notify clients about a value change. Sends the value."""
+        self.broker.publish(self.value, self)
+
+    def to_dict(self):
```
thomaspurchas: Not sure about this. See comment in PR.

pyhap/characteristic.py
```diff
@@ -189,50 +151,44 @@ def override_properties(self, properties=None, valid_values=None):
...
-        hap_rep.update(value_info)
-        return hap_rep
+        Change self.value to value and call callback.
+        The callback method must return a boolean that indicates if successful.
+        """
+        self.value = value
+        self.notify()
+        if self.setter_callback:
+            return self.setter_callback(value)
+        return True
```
thomaspurchas: Rather than returning `True`/`False` I think the callback should raise an exception on failure. The exception could be different depending on what went wrong.

We can then map the exceptions to the HAP protocol status values.

Using `True`/`False` feels a little bit like err values in C.

pyhap/characteristic.py
```diff
@@ -189,50 +151,44 @@ def override_properties(self, properties=None, valid_values=None):
...
-            value_info["value"] = self.value
+    def update_value(self, value):
+        """Called from broker for value change in Home app.
 
-        hap_rep.update(value_info)
-        return hap_rep
+        Change self.value to value and call callback.
+        The callback method must return a boolean that indicates if successful.
+        """
+        self.value = value
```
thomaspurchas: Value should only be set after the `setter_callback` is successful, otherwise the `value` won't represent what the device did.

pyhap/characteristic.py
```diff
@@ -189,50 +151,44 @@ def override_properties(self, properties=None, valid_values=None):
...
+        self.notify()
+        if self.setter_callback:
+            return self.setter_callback(value)
+        return True
+
+    @classmethod
+    def from_dict(cls, name, json_dict):
```
thomaspurchas: When do you imagine this will be used?

pyhap/accessories/SDS011.py
cdce8p: I didn't mentioned that above, but I realized that there are cases in which setting the value on its one does make sense. Testing as well as the one above. `value = 0` is basically equivalent to `set_value(0, should_notify=False)` if no callback has been added and the char doesn't belong to any service / accessory yet.

pyhap/characteristic.py
ikalchev: I would prefer `to_HAP` for the same reasons to be honest. If `to_dict` is so needed, it can currently just return the result of `to_HAP`.

cdce8p: > to_HAP exists to create a representation of the characteristic that the HAP protocol understands. The fact that it's a dict is just an implementation detail, and it shouldn't be relied on.

I disagree. It might just be an implementation detail, but that exactly makes it even better. `to_HAP` might not be clear for someone without a background in HomeKit who just started looking at `HAP-python`.

> Changing it to to_dict give the impression that you could use that representation for something else

Why not? If someone whats to use that say for debugging, why should we limit that?

> Changing it to to_dict give the impression [...] that it's always going to be a dict type. Neither of those things are true, in the future we may decide to introduce a HAPMessage type or something similar, which would be returned instead of a dict. to_dict ties our hands on that, conflating what the function does with what it returns.

I get that, but what prevents us to changing that back in the future?

---

I believe that `to_dict` would be a better representation of this method since this method name is widely used across many different python projects, including Home Assistant (for `States` and `Events`).
However if you two think otherwise, this shouldn't stand in the way of this PR.

pyhap/characteristic.py
cdce8p: I would think that `True / False` should be enough for our proposes, but you are right in that exceptions would allow for more options. @ikalchev What do you think?

pyhap/characteristic.py
```diff
@@ -91,60 +89,49 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
...
-        return "<characteristic display_name='{}' value={} properties={}>" \
+        return '<characteristic display_name={} value={} properties={}>' \
             .format(self.display_name, self.value, self.properties)
 
-    def set_value(self, value, should_notify=True, should_callback=True):
-        """Set the given raw value. It is checked if it is a valid value.
-
-        :param value: The value to assign as this Characteristic's value.
-        :type value: Depends on properties["Format"]
+    def set_value(self, value):
```
ikalchev: I am still thinking about the `should_notify` and if there is a case for it. Image you want to set the value but not notify iOS clients - technically you can do it by setting the value property but then you don't have the `to_valid_value` transformation. What do you think, is it worth adding it?

pyhap/characteristic.py
```diff
@@ -163,70 +150,45 @@ def override_properties(self, properties=None, valid_values=None):
...
+                json_dict['maxLen'] = min(len(self.value), 256)
+        if HAP_PERMISSIONS.READ in self.properties['Permissions']:
+            json_dict['value'] = self.value
 
-        .. note:: Uses the `get_hap_value`, i.e. sends the HAP value.
+        return json_dict
 
-        :param iid_manager: IID manager to query for this object's IID.
-        :type iid_manager: IIDManager
+    def update_value(self, value):
```
ikalchev: How about `client_set_value` to make it more clear?

cdce8p: Sounds good :+1:

pyhap/accessories/SDS011.py
thomaspurchas: Yes, but doing this will ignore all the validation logic, which still needs to be run at some point. This could allow an invalid value to be added using `__init__` and then sent to an iOS client during a `get`.

Better to just use `set_value`, and have the char figure out if sending a notification makes sense (if there is no broker, then just don't bother with the notification).

pyhap/characteristic.py
ikalchev: You can use `to_HAP` for debugging without it being `to_dict`. How about this:
```python
def to_dict(self):
    return self.to_HAP()
```
In this way, we get best of both worlds. What is more, when/if `to_dict` changes because there are more stuff in the Char we want to dict, we can just add those in `to_dict`

thomaspurchas: Still not convinced by this. `to_HAP` has a very explicit purpose, which is to serialise the object for the HAP proto. 

I agree with @ikalchev, either `to_dict` should return the result of `to_HAP`, or perhaps better, `to_HAP` uses `to_dict` internally to build its serialisation. 

Either way `to_HAP` shouldn't go anywhere as it's an important internal API. But I don't see any issues with added an additional `to_dict` method.

pyhap/characteristic.py
cdce8p: I though about this case as well. It's a trade of here. I think that e.g. `value = 0` would be enough, since most of those cases are for static content which is hard coded. Therefore it would be quite easy to recognize a problem / bug and change it.

pyhap/characteristic.py
thomaspurchas: I would point out that `True`/`False` is not backwards compatible, and requires that users remember to return a `True` value on success (forgetting to return a value would result in `None`). This does not strike me as very pythonic.

pyhap/characteristic.py
thomaspurchas: Yeah adding a `should_notify=True` would probably be a good move.

pyhap/characteristic.py
ikalchev: I just looked at Apple's spec - there are 11 error codes (which I need to add btw). If we use exceptions, we will force the Driver from understanding these exceptions and mapping them to error codes - in this case, why not `set_value` to return one of the error codes instead? What is more, returning an error code would allow "internal" errors to more "voicefully" propagate with an exception.

pyhap/characteristic.py
thomaspurchas: Could I suggest `client_change_value`, I think it makes it clearer that I might fail. But I could just be bikeshedding.

pyhap/characteristic.py
cdce8p: I will leave it at `to_HAP`.
```py
def to_dict(self):
    return self.to_HAP()
```
That would add just another method to do the same think and would defeat the propose of this change (to reduce complexity). I didn't thought that would be so controversial :sweat_smile:

pyhap/characteristic.py
thomaspurchas: There might be 11 error codes, but only some would be valid for a callback i.e. `Request denied due to insufficient privileges.` `Unable to communicate with requested service, e.g. the power to the accessory was turned off.` `Resource is busy, try again.` `Accessory received an invalid value in a write request.`

The rest are errors that should be managed by py-HAP. i.e. `Operation timed out.` could be thrown by setting a timeout on the callback (either via threads or asyncio).

ikalchev: True, but let's keep in mind that tomorrow Apple may add 10 more, 4 of which would make sense (for example).

cdce8p: What do you guys think if I exclude the error handling here and just call the callback method. That way we can implement it in another PR together with the complete error handling.

pyhap/characteristic.py
thomaspurchas: To nit pick, I think it actually makes more sense to do it the other way around. `to_HAP` calls `to_dict` and then trims the result.

ikalchev: What one developer can do in one month, two developers can do in two months.

ikalchev: > to_HAP calls to_dict and then trims the result.
Probably better, yes

pyhap/characteristic.py
ikalchev: Sounds good :)

thomaspurchas: @ikalchev still doesn't feel pythonic. We would then need a bunch of static vars somewhere to match the docs, or rely on users parsing the docs.

Not sure how best to move this convo into another PR.

cdce8p: @thomaspurchas maybe continue here: #75

ikalchev: Let's think on that, it's not such a big deal, whatever we chose.

BTW there is a discord chat, but I haven't mentioned that here because it would make things more hard to follow.

pyhap/characteristic.py
cdce8p: I plan to use it for the `Char_Loader` class, but it might make more sense if added when there is a use for it. I will remove it for now.

pyhap/characteristic.py
thomaspurchas: I've copied a summary of this thread into there.

pyhap/characteristic.py
cdce8p: I guess it wont hurt anyone with `should_notify=True`

pyhap/characteristic.py
cdce8p: For a representation / debugging: `__repr__` would be the way to go. I would stick with `to_HAP` for now then.

pyhap/characteristic.py
cdce8p: What about `client_update_value`?

pyhap/characteristic.py
```diff
@@ -4,6 +4,10 @@
 A Characteristic is the smallest unit of the smart home, e.g.
 a temperature measuring or a device status.
 """
+import logging
+
+_LOGGER = logging.getLogger(__name__)
```
ikalchev: I know this is pythonic and all that but using `logger` instead is also fairly common and easier to write. What do you think?

pyhap/characteristic.py
```diff
@@ -91,61 +84,57 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
...
+            return min(self.properties['ValidValues'].values())
+        else:
+            value = HAP_FORMAT.DEFAULT[self.properties['Format']] 
+            return self.to_valid_value(value)
+
+    def to_valid_value(self, value):
+        """Perform validation and conversion to valid value"""
+        if self.properties.get('ValidValues'):
+            if value not in self.properties['ValidValues'].values():
+                raise ValueError('%s is not a valid value'.format(value))
```
cdce8p: @ikalchev I was thinking about this. What would you think about an error log statement instead of an ValueError? The log statement would be easier to debug.

pyhap/characteristic.py
cdce8p: That's what I would have suggested as well. I will change it to `logger`.

pyhap/characteristic.py
ikalchev: Why do you think so?

cdce8p: With errors you get the whole traceback and it might seem unintentional. Whereas with error log message this would be handled much cleaner.

I started thinking in which cases you would use an `raise Error`. It is only if you don't know how to properly handle it in your program. So you push the responsibility to the end user.

As an implementation for us: I thought about `logger.error` and `return None`. Adding an `check if is not None` to `set_value`, otherwise abort.

ikalchev: How about `logger.error`, `return value # as is`. iOS will show no response or something if it gets an invalid value ( I think)

cdce8p: I think that breaks HomeKit. I tried it some time ago and it seemed to cause issues. I believe if the value doesn't get updated, that should be hint enough to think something is wrong here.

cdce8p:
If you're ok with these, I would like to rearrange the functions before merging.
I would still have to update the test suite though.

pyhap/characteristic.py
thomaspurchas: Why would you not want an Exception and traceback? If the value is invalid then the caller should capture it, and either log it or ignore it (py-HAP is just providing an API after all, not then entire application. The demo server could be updated to capture and log the exception).

If the iOS client makes a call and sets an invalid value, then the exception should also be raised, caught by `client_set_value()` and the appropriate status response sent back to iOS.

pyhap/characteristic.py
```diff
@@ -125,12 +130,18 @@ def to_valid_value(self, value):
         """Perform validation and conversion to valid value"""
         if self.properties.get('ValidValues'):
             if value not in self.properties['ValidValues'].values():
-                raise ValueError('%s is not a valid value'.format(value))
+                logger.error('%s: value=%s is an invalid value.',
```
thomaspurchas: Sorry, didn't my comment in quick enough before.

It seems really strange to handle the error here. Surely it's better just to raise an Exception and have the caller handle it. Doing this means we now have special return values to indicate a failure.

If people don't like the traceback, then they should be catching the `ValueError` and logging it explicitly. (We could modify the examples to do this).

Addtionally if the `ValueError` is caused by iOS, then we should be returning the correct status code to iOS to indicate an invalid value has been sent.

pyhap/characteristic.py
cdce8p: I don't think we should had over the responsibility for that.  Especially because we would need a try block for every `set_value` call.

For `client_set_value()`: By definition are all values stored in `self.value` are valid values for HomeKit. So HomeKit can't set an invalid value to `HAP-python`. There is just the case were it's an invalid value for the object / sensor you control, but I think this is for #75

pyhap/characteristic.py
thomaspurchas: I wouldn't catch in `set_value` but in whatever called `set_value`. It seems strange to rob the caller of the ability to handle the error, just because our one usage now dosen't require that ability, and tracebacks are ugly.

thomaspurchas: Also returning `None` hides the error from the caller. Making even less obvious to the developer using the API that something has gone wrong.

Submitting an invalid value the perfect time to fail fast and fail hard. Ensuring that anything that might rely on that valuing be set won't run unless the failure is explicitly handled. 

Doing anything else if just going to quietly hide the error in impenetrable logs.

cdce8p: I'm thinking from a general standpoint here. You should configure your accessory the way that only valid values are set. If for some reason a value isn't valid, you should get a notification with the invalid value so you can update your program.
It would be a waist of resources to write the error handling for every set_value call individually. Think of the Home Assistant `Thermostat` implementation. It alone uses 11 set value calls.

thomaspurchas: By your own argument then, there is no reason not to raise an exception. If they only happen rarely, and always indicate that a serious problem has occurred, then they should just be allowed to peculate to the top. Doesn't matter how the traceback looks, it should never happen.

You don't need to do error handling on every call to `set_value`, only on calls where you are pulling values from somewhere that you don't control, and thus can't guarantee will always be in bounds.

Additionally we can add some smarter error handling at the top of the accessory to capture these exceptions, and restart the appropriate accessory loop if people want that.

cdce8p: Go ahead, but I'm frustrated if I see those error logs and should try to find the issue: https://github.com/home-assistant/home-assistant/issues/13676

At least for Home Assistant you can't never be sure that the correct values are supplied and there will always be one you haven't caught.

thomaspurchas: I understand that, but the solution isn't to remove functionality from py-HAP. We should modify the accessory components when we integrate with HA to capture the exceptions. This could simply be done by creating a `set_value_log_errors()` helper in HA, or on the accessory.

This way we also get the added benefit of being able to write HA specific messages, which can pull in `entity_id` and component name. Giving us the debug info needed to fix components.

cdce8p: Of course that is possible. I just think it defies the purpose. A user shouldn't ever get tracebacks. If an error occurs it should be handled as error log message.

cdce8p: It would be different if `HAP-python` would just be an API the user doesn't interact with directly, but it isn't.

thomaspurchas: I agree that a user shouldn’t get trace backs. But it’s on HA to catch Exceptions, it’s not for py-HAP to accept any old input and carry on, just so HA doesn’t have do proper error handling. 

py-HAP is just an API, a HA user shouldn’t have to interact directly with it. The HA ui should be doing all the heavy lifting. 

If people are using py-HAP directly then it’s on them to catch errors and handle them, and they should be devs and know what to do with them. 

If people are just using the demo accessories then they should be updated to capture the exception and produce a meaningful error message. 

In each these scenarios the output will be different, we should be picking one and forcing everyone to use it. 

And specifically with HA, doing the error handling in HA, rather than in py-hap means we can produce much better error messages(including HA info). Helping us debug issues quicker.

thomaspurchas: Can I also point out that your proposed error message is very unhelpful. It says there is an invalid value, be not what caused it. Which accessory tried to set the value, and onto what characteristic, and why is it invalid?

thomaspurchas: One final point. If your frustrated by the amount of work needed to do the error handling. I’m happy to do that work, both here and in HA.

cdce8p: I disagree. `'%s: value=%s is an invalid value.', self.display_name, value` should cover almost all cases. It's true that you don't get the accessory, but you know it already since it hasen't been updated. You also get the distinction between `valid values` and `numerics`. There isn't really anything more I would add / need.

---

> One final point. If your frustrated by the amount of work needed to do the error handling. I’m happy to do that work, both here and in HA.

I won't stop you from looking through the open issues and suggesting solutions.

thomaspurchas:
Very well, my first suggestion is to leave this exception in.

For now, all I ask is you leave this commit out of this PR. Tomorrow I’ll go though and create PRs for better error handling in HA, and in the demo accessories in py-HAP.

Once I’ve done the PRs, if you still think dropping this exception is the best move, then we’ll do that move onto other issues.

cdce8p:
We don't get anywhere here. @ikalchev What do you think?

pyhap/characteristic.py
ikalchev: I think we are overthinking this a bit. It should be very rare to get invalid values:
- iOS cannot send an invalid value, because it picks values from the min/max/validValues fields that the characteristic defines. In the worst case, even if Apple change something, your accessory will not get added to the Home app in the first place (I tried)
- Accessories in HAP are developed from devs not end users. This is the target "audience" so to speak. Thus, we need the best solution for devs to be able to pinpoint the problem.

That being said, I think that the current ValueError is good enough. When the dev is coding a new Accessory, he/she will probably get these early in testing. Once they are gone, they should be gone for good.

cdce8p:
I was testing the changes and found something interesting: At least in Home Assistant the Value Error will not be displayed by default until it is stopped. That might be true for other implementations as well.
I would suggest adding `logger.error` additionally before `raise ValueError` to have at least the the error message during runtime.

ikalchev:
This is odd. Is that HA problem or by design?

pyhap/characteristic.py
```diff
@@ -91,61 +84,68 @@ def __init__(self, display_name, type_id, properties, value=None, broker=None):
...
+            return min(self.properties['ValidValues'].values())
+        else:
+            value = HAP_FORMAT.DEFAULT[self.properties['Format']] 
+            return self.to_valid_value(value)
+
+    def to_valid_value(self, value):
+        """Perform validation and conversion to valid value"""
+        if self.properties.get('ValidValues'):
+            if value not in self.properties['ValidValues'].values():
+                logger.error('%s: value={} is an invalid value.',
```
ikalchev: I don’t want to nit pick, but can we `e = str.format(...); logger.error(e); raise ValueError(e)`

cdce8p: I named the string `error_msg` if that's ok. New commit is online

cdce8p:
> This is odd. Is that HA problem or by design?

I'm not sure. I did get the Value error once right after the call to set_value. But the was out of over ten tries.

cdce8p:
Depends if I should update the tests. Feature wise it's ready.

ikalchev:
As long as they are passing, I think we are ok

cdce8p:
They are not currently, since there are breaking changes.

ikalchev:
Would it be OK if you try to fix them? Alternatively, If you are sick of this change I don't mind doing it

cdce8p:
I can do them. Just won't get to it until in a few hours, since I'm currently away from my computer. I  should get them ready today.

ikalchev:
No hurry :)

thomaspurchas:
>>> I was testing the changes and found something interesting: At least in Home Assistant the Value Error will not be displayed by default until it is stopped. That might be true for other implementations as well.
>>> I would suggest adding logger.error additionally before raise ValueError to have at least the the error message during runtime.
>> 
>>  This is odd. Is that HA problem or by design?
> 
> I'm not sure. I did get the Value error once right after the call to set_value. But the was out of over ten tries.

This will be an artefact of asyncio. The exception is being raised, and capture by a `Future` for inspection later, and that `Future` is only being cleaned up on the event loop shutdown. Also behaviou might change depending on asyncio debug settings.

Relevant python docs here: https://docs.python.org/3/library/asyncio-dev.html#detect-exceptions-never-consumed

Also worth looking at this: https://stackoverflow.com/questions/30361824/asynchronous-exception-handling-in-python

cdce8p:
I just finished the tests. Should be 100% coverage, although that doesn't mean much.
@ikalchev If I haven't forgotten anything, this should be it. Ready to merge

ikalchev:
This is a lot of great work! Cheers!

Issue #64, opened 2018-03-22
Async support


cdce8p:
Currently `HAP-python` uses threads which, at least on a larger scale, does impact performance quite a bit. Are there plans to change to an async framework, like `asyncio` in the future?

Unfortunately this would also require a more recent min python version, to take full advantage of the new async syntax introduced with py3.5.

ikalchev:
This is actually a very good idea, which I think we might pursue. I was recently thinking on the threads starting to be an issue and it looks like asyncio can fit the bill nicely.

Dropping 3.4 will not be a huge issue I believe. I had similar requests before, but no good reasons to do it.

I am currently busier at work than what is usual, but I expect things to be back to normal from this weekend onwards. I will definitely spend some time on this soon and try to make some first steps. Of course, you are welcome to do further suggestions and PRs if you want.

On a side note, @jslay88 started a chat for the project on discord I think recently which I hadn't had the time to "take over" yet. I will try to kick this off as well.

Best,
Ivan

jslay88:
https://discord.gg/BhxZ4Dy

So lonely :)

cdce8p:
Good to here that. I'm happy to help, although I just started watching talks and tutorials about this two weeks ago :sweat_smile: I didn't wanted to start before knowing if this is something you want. It's quite some work and will require major changes to the code base.

Maybe we can coordinate efforts on the discord channel than

ikalchev:
FYI started transforming the Accessory/Bridge run methods to use asyncio (seemed easiest starting point)

cdce8p:
:+1: Could you open a `[WIP] PR` for it?

ikalchev:
Yes, good point, I am doing this now since I moved to dev and master branches. I hadn't had anything significant to push recently due to lack of time.

ikalchev:
@cdce8p have you used aiohttp https://aiohttp.readthedocs.io/en/stable/

I am considering it as the next step (after the Accessories run method).

cdce8p:
@ikalchev Not yet, since I didn't have/had any experience with networking prior to working with `HAP-python`. `aiohttp` seems to be the right choice though.

Did you open a WIP PR? That would really help. Especially to know what will be changed and to give suggestions.

cdce8p:
Work in progress

cdce8p:
[...]

PR #74, opened 2018-03-31
[WIP] Move to coroutines for Accessory.run

ikalchev:
Problems to solve:
- Support non-coroutine run methods - not all code could switch to
asyncio, these should be started in a thread as before.
- Port accessories/*

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -381,6 +381,28 @@ class Accessory(object):
         self.print_qr()
         print('Or enter this code in your HomeKit app on your iOS device: %s' % self.pincode.decode())
 
+    def run_at_interval(seconds):
+        """Decorator that runs decorated method in a while loop, which repeats every
+        ``seconds`` until the ``Accessory.run_sentinel`` is set.
+
+        .. code-block:: python
+
+            @Accessory.run_at_interval(3)
+            def run(self):
+                print("Hello again world!")
+
+        :param seconds: The amount of seconds to wait for the event to be set.
+            Determines the interval on which the decorated method will be called.
+        :type seconds: float
+        """
+        # decorator returns a decorator with the argument it got
+        def _repeat(func):
+            def _wrapper(self, *args, **kwargs):
+                while not self.run_sentinel.wait(seconds):
+                    func(self, *args, **kwargs)
+            return _wrapper
+        return _repeat
+
     def run(self):
         """Called when the Accessory should start doing its thing.
 
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -381,41 +381,12 @@ class Accessory(object):
         self.print_qr()
         print('Or enter this code in your HomeKit app on your iOS device: %s' % self.pincode.decode())
 
-    def repeat(seconds):
-        """Decorator that runs decorated method in a while loop, which repeats every
-        ``seconds``. It ues the ``Accessory.aio_stop_event``.
-
-        :param seconds: The amount of seconds to wait for the event to be set.
-            Determines the interval on which the decorated method will be called.
-        :type seconds: float
-        """
-        def _repeat(func):
-            async def _wrapper(self, *args, **kwargs):
-                while not await util.event_wait(self.aio_stop_event, seconds):
-                    await func(self, *args, **kwargs)
-            return _wrapper
-        return _repeat
-
-    async def run(self):
+    def run(self):
         """Called when the Accessory should start doing its thing.
 
         Called when HAP server is running, advertising is set, etc.
         """
         pass
-
-    async def _wrap_in_thread(self):
-        threading.Thread(target=self.run).start()
-
-    async def start(self):
-        """Create and await on a task for ``Accessory.run``
-
-        If ``Accessory.run`` is not a coroutine, it will be wrapped in a task that
-        starts it in a new thread.
-        """
-        if asyncio.iscoroutinefunction(self.run):
-            await self.event_loop.create_task(self.run())
-        else:
-            await self.event_loop.create_task(self._wrap_in_thread())
 
     def stop(self):
         """Called when the Accessory should stop what is doing and clean up any resources.
@@ -448,7 +419,40 @@ class Accessory(object):
         self.broker.publish(acc_data)
 
 
-class Bridge(Accessory):
+class AsyncAccessory(Accessory):
+
+    def run_at_interval(seconds):
+        """Decorator that runs decorated method in a while loop, which repeats every
+        ``seconds`` until the ``Accessory.aio_stop_event`` is set.
+
+        Can be used as:
+        .. code-block:: python
+
+        @AsyncAccessory.run_at_interval(3)
+        async def run(self):
+            print("Hello again world!")
+
+        :param seconds: The amount of seconds to wait for the event to be set.
+            Determines the interval on which the decorated method will be called.
+        :type seconds: float
+        """
+        # decorator returns a decorator with the argument it got
+        def _repeat(func):
+            async def _wrapper(self, *args, **kwargs):
+                while not await util.event_wait(self.aio_stop_event,
+                                                seconds,
+                                                self.event_loop):
+                    await func(self, *args, **kwargs)
+            return _wrapper
+        return _repeat
+
+    async def run(self):
+        """Override in the implementation if needed.
+        """
+        pass
+
+
+class Bridge(AsyncAccessory):
     """A representation of a HAP bridge.
 
     A `Bridge` can have multiple `Accessories`.
@@ -469,7 +473,7 @@ class Bridge(Accessory):
 
     def set_sentinel(self, run_sentinel, aio_stop_event, event_loop):
         """Set the same sentinel to all contained accessories."""
-        super(Bridge, self).set_sentinel(run_sentinel, aio_stop_event, event_loop)
+        super().set_sentinel(run_sentinel, aio_stop_event, event_loop)
         for acc in self.accessories.values():
             acc.set_sentinel(run_sentinel, aio_stop_event, event_loop)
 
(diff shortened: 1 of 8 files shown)


ikalchev:
First steps for #64

thomaspurchas:
Do you have a plan of work for this? Happy to pitch in, just don't want to end up doing the same work twice.

cdce8p:
I'm by far an expert on async. Quite a while ago Home Assistant converted their core to async and I think that might be a good help to take a look at: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/core.py

Some takeaways they do:
 - The basis in async and everything that still uses threads is run through a `ThreadPoolExecutor`
 - Helper methods to append methods / coroutines to the loop / executer, without the need to pass loop to every function

ikalchev:
@cdce8p this was insightful, thanks. It's pretty similar to where I was heading, except the pool, which is much cleaner.

@thomaspurchas first let me briefly describe the threads in HAP-python and their purpose:
- The server thread - this listens for iOS clients and spawns a new thread to handle each new connection.
- The accessory threads - This is either one thread or multiple threads, depending on if there is a single accessory or a bridge containing multiple accessories. Each of these runs the `Accessory.run`. Eventually, value updates (such as reading the temperature from a sensor) result in an event being sent to iOS clients. This is achieved by each thread writing the event to a thread-safe queue (the event queue).
- The "send events" thread - Because we don't want the Accessory threads to block waiting for the event to be sent, there is another thread which reads from the event queue and sends updates to clients. This also handles stale connections, etc.

The biggest "thread overhead" comes from the accessory threads because you have N threads for N accessories. What is more, they mostly sleep and do GPIO (e.g. read the sensor every X minutes). So if I have to outline a roadmap for this:

- [x] Move from accessory threads to a mix of coroutines and threads.
- [ ] Evaluate if moving to [aiohttp](https://aiohttp.readthedocs.io/en/stable/) is worth it. This is not clear, because each client retains an open two-way connection and I am not entirely sure that moving to asyncio will be better. Some factors are the average number of clients and the average frequency of events coming from the accessories.

thomaspurchas:
Could you elaborate a bit more on:

> Evaluate if moving to aiohttp is worth it. This is not clear, because each client retains an open two-way connection and I am not entirely sure that moving to asyncio will be better. Some factors are the average number of clients and the average frequency of events coming from the accessories

If asyncio is anything like Twisted (which did async stuff and co-routines in Python a decade ago), then using aiohttp will still be worth it. The connection might stay open, but you hand-off waiting for data to the event loop, which will sit on network file-descriptor using `select` or `poll`/`epoll` waiting for data. So you code will only be called when data is moving about. Using asyncio everywhere will prevent botched code where we need to hand data between the event loop, and threads, which is never nice.

I'll do a little research to get a better idea of how hard if will be to do long-polling in aiohttp. I would agree though, that moving to aiohttp is probably lower priority.

thomaspurchas:
This stuff will have an impact on #75, it would be cool if callbacks could be co-routines. That would break API compatibility though, however would result in much cleaner code in the long run.

thomaspurchas:
After doing the little work I did. I realised that it might be a good move to split the `Accessory` class into an `AsyncAccessory` and `SyncAccessory`.

That would make it much easier to reason about the code because their would be clear separation. Then the rest of HAP-Python can be rewritten with async support everywhere, and older stuff can just use the `SyncAccessory` class for backwards compatibility.

pyhap/accessory_driver.py
```diff
@@ -96,6 +99,32 @@ class HAP_CONSTANTS:
...
+    controlling the Accessory.run method.
+    """
+
+    def __init__(self, run_method):
+        """
+        """
+        self.loop = asyncio.new_event_loop()
+        self.stop_event = asyncio.Event()
+        self.task = self.loop.create_task(run_method(self.stop_event, self.loop))
+        super(AIOThread, self).__init__(target=self.loop.run_until_complete,
```
thomaspurchas: There is no need for `super(AIOThread, self).__init__(`, this can be shorted to `super().__init__(` in Python 3 and higher.

pyhap/accessory_driver.py
```diff
@@ -96,6 +99,32 @@ class HAP_CONSTANTS:
...
+        try:
+            super(AIOThread, self).run()
+        except CancelledError:
+            self.loop.stop()
+            self.loop.close()
+            logger.info("Sucessfully stopped accessory event loop.")
+
+    def stop(self):
+        self.loop.call_soon_threadsafe(self.task.cancel)
+
```
thomaspurchas: I don't think this class is safe. It looks like you are using a single event loop from multiple threads here. I'm 99% sure you shouldn't do that. (I've had exceptions thrown as a result).

Instead the event loop should be setup after thread creation like this:

```python
class AIOThread(threading.Thread):
    """
    TODO: temporary convenience class. Create an event loop and event object for
    controlling the Accessory.run method.
    """

    def __init__(self, run_method):
        """
        """
        self.run_method = run_method
        super().__init__()

    def run(self):
        self.loop = asyncio.new_event_loop()
        asyncio.set_event_loop(self.loop)

        self.stop_event = asyncio.Event()
        self.task = self.loop.create_task(self.run_method(self.stop_event, self.loop))

        self.loop.run_until_complete(self.task)

    async def safe_shutdown(self):
        try:
            self.stop_event.set()
            await asyncio.sleep(5)
            if not self.task.done():
                self.task.cancel()
                await asyncio.sleep(5)
            self.loop.stop()
        finally:
            self.loop.close()
            logger.info("Sucessfully stopped accessory event loop.")

    def stop(self):
        self.loop.call_soon_threadsafe(self.safe_shutdown)
```

pyhap/accessory.py
```diff
@@ -377,13 +381,42 @@ def setup_message(self):
         self.print_qr()
         print('Or enter this code in your HomeKit app on your iOS device: %s' % self.pincode.decode())
 
-    async def run(self, stop_event, loop=None):
+    def repeat(seconds):
```
thomaspurchas: Maybe rename `every` or `interval`. `repeat` give the impression the decorated function will be called 3 times, then stop.

pyhap/accessory.py
```diff
@@ -434,11 +467,11 @@ def __init__(self, display_name, mac=None, pincode=None,
                                      setup_id=setup_id)
         self.accessories = {}  # aid: acc
 
-    def set_sentinel(self, run_sentinel):
+    def set_sentinel(self, run_sentinel, aio_stop_event, event_loop):
         """Set the same sentinel to all contained accessories."""
-        super(Bridge, self).set_sentinel(run_sentinel)
+        super(Bridge, self).set_sentinel(run_sentinel, aio_stop_event, event_loop)
```
thomaspurchas: No need for `super(Bridge, self).set_sentinel()`. `super().set_sentinel` is all that is needed.

pyhap/accessory_driver.py
```diff
@@ -519,6 +491,15 @@ def start(self):
         if not self.accessory.paired:
             self.accessory.setup_message()
 
+        # Start the accessory so it can do stuff.
+        self.accessory.set_sentinel(self.stop_event, self.aio_stop_event, self.event_loop)
+        self.accessory_task = self.event_loop.create_task(self.accessory.start())
+        logger.info("Starting event loop")
+        self.event_loop.run_until_complete(self.accessory_task)
+        self.event_loop.stop()
```
thomaspurchas: No need to call `stop`. `run_until_complete` already does this for you. (In fact, `run_until_complete` can only return once the loop has stopped)

ikalchev:
Hi all,

I posted a new version and I think I am almost happy with it. There a few things I need to address, but overall I think it ramps up to be a nice solution. I want to hear what's on your mind, if you wish, so go ahead.

A summary will be:
- Don't break users already using the library.
- Thus, we want both threads and coroutines from a single interface
- This single interface is `Accessory.start` for the Driver side, while we keep `Accessory.run` as the user side.
- `Accessory.run` can be both sync and async. In the former case, `Accessory.start` wraps that in a coroutine that starts a thread, so that we can iterate over multiple Accessory objects and schedule their run methods as tasks.
- To make all of this work and avoid using `task.cancel`, I added `event_wait` in the `util` module
- There is also a fancy decorator that allows new users to drop the use of `while not event.set..` and just do:
```
@Accessory.repeat(3)
async def run(self):
    print("Hello world")  # prints every 3 seconds
```
- Finally, to synchronise all these I pass a `threading.Event`, a `asyncio.Event` and an event loop to the Accessories. This is the part I would like to refine a bit.

In the meantime, I will try to setup a new async branch as suggested and go trough all the PRs and comments as soon as I can, sorry for the delay.

pyhap/accessory_driver.py
```diff
@@ -185,6 +159,9 @@ def __init__(self, accessory, port, address=None, persist_file="accessory.state"
             self.persist()
         self.topics = {}  # topic: set of (address, port) of subscribed clients
         self.topic_lock = threading.Lock()  # for exclusive access to the topics
+        self.event_loop = asyncio.get_event_loop()
+        self.aio_stop_event = asyncio.Event()
```
thomaspurchas: I think there is a better, cleaner way to do this.

It should be possible to call `self.stop_event` using `event_loop.run_in_executor()` to wait on the threaded event in a separate thread managed by asyncio. Then put a callback on the returned future to call `aio_stop_event.set()` back in the main thread.

That way calling `stop_event.set()` anywhere is both threadsafe, and guaranteed to both threads and the `event_loop`.

I'll put together a PR once I figured out the nicest way the structure the code.

thomaspurchas:
With regards to:

> - This single interface is Accessory.start for the Driver side, while we keep Accessory.run as the user side.
> - Accessory.run can be both sync and async. In the former case, Accessory.start wraps that in a coroutine that starts a thread, so that we can iterate over multiple Accessory objects and schedule their run methods as tasks.

I think it might be an idea to move the async stuff into a separate `AsyncAccessory` class. Because more than `run` should be a co-routines. Ideally any char callbacks should also be co-routines.

This way we can encourage either 100% sync or 100% async accessories, and prevent accessories mixing both approaches (which would be nightmare from a execution safely perspective).

pyhap/accessory_driver.py
ikalchev: I agree that it will be better. Do you mind if we leave it for another change?

pyhap/accessory.py
ikalchev: Addressed.

pyhap/accessory_driver.py
ikalchev: Done

pyhap/accessory.py
ikalchev: Agree. `run_at_interval`?

ikalchev:
Hopefully a fancier change

Issue #43, opened 2018-02-18
Suggestion: HAP-python-min


cdce8p:
I was thinking if it would make sense to add a new **pypi** project: `HAP-python-min`. This would include all content from `pyhap` except for the `accessories` folder.

I'm currently working on a component for [Home Assistant](https://github.com/home-assistant/home-assistant). Some of the examples form the accessories folder may work, but to port of all functions is basically more difficult than to write a new class which extends `accessory`. Because the initial code base for Home Assistant is already big their is a focus to keep additional imports as small as possible to improve performance. A `min` version would help accomplish that.

schinckel:
I was thinking about this too: whether we package up third-party accessories elsewhere. Perhaps we should/could create a HAP-python team or whatever GitHub calls them, and put packages into there. Could make it easier to have new packages/authors involved, too.

ikalchev:
Okay I will check how to do that, any sharing of know-how is appreciated.

jslay88:
I like this idea about the team and moving accessories elsewhere. I have some pretty specific ones that I would like to include, but feel they are too specific for the main repo.

ikalchev:
Will work on this now.

ikalchev:
For Option B, things will look like this
```
pyhap/
   __init__.py
   ...
   accessories/
      # no __init__.py here; implicit namespace package
      fakes/
         # fake accessories here, e.g. FakeFan, TemperatureSensor
         __init__.py
      sensors/
         # e.g. accessories that measure stuff
         __init__.py
      ...
```
Third-parties can add sub packages to accessories following this directory structure in their repo.

See here https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages.

cdce8p:
I'm no expert on packaging. However I don't think Option C will achive the objectives of this change.

Besides reducing the package size, extracting the accessories to their own repo would make development easier since we can update them for a new version seperately.

Therefore I would prefer either `A` or `B`.

ikalchev:
Done. See `pyhap.accessories.README.md` for integrating as a sub package.

PR #76, opened 2018-04-01
[WIP] Port accessories to asyncio

thomaspurchas:
This should merge into #74. Gonna start working on moving the HTTP bridge.

Unfortunately there are some limits to running the event loop in threads. Notably subprocess don't work in threads, [unless the main thread is running an event loop](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio-subprocess-threads).

I think I might start work migrating the driver over to asyncio, so an event loop is running in the main thread.

With regards to this port, feedback welcome, also testing as I don't have access to all the hardware.

### DIFF
--- a/pyhap/accessories/AM2302.py
+++ b/pyhap/accessories/AM2302.py
@@ -4,6 +4,7 @@ Also, make sure pigpiod is running.
 The DHT22 module was taken from
 https://www.raspberrypi.org/forums/viewtopic.php?f=37&t=71336
 """
+import asyncio
 import time
 import random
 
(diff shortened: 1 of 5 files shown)


ikalchev:
Hey! I will go over the PR later when I get to a computer. As for the loop in the driver - this is the core of the PR I opened, would you mind if I finish it?

thomaspurchas:
Ahh, I thought the PR you opened was only for the `Accessory.run` method. Didn't realise you where gonna redo the accessory driver as well!

ikalchev:
Hey! Too busy the last week, but I submitted the asyncio today (in dev). Could you revise this?

requirements_all.txt
```diff
@@ -4,4 +4,5 @@ ed25519
 zeroconf
 curve25519-donna
 pyqrcode
-base36
\ No newline at end of file
+base36
+aiohttp==3.1.1
```
cdce8p: I don't like that we add a dependency that is only used for an optional accessory. If that's the way to go, we should move all optional accessories elsewhere #43.

ikalchev:
Note that one of the merges broke the accessories (the Category is now in costs.py and is not a class). Will try to fix this soon

PR #81, opened 2018-04-11
Move scripts to separate folder

cdce8p:
If scripts are executed from `HAP-python` folder nothing should change.
E.g.:
```bash
$ cd HAP-python
$ script/gen_hap_types.py
```


ikalchev:
Looks good, can we just change the name to scripts?

cdce8p:
> Looks good, can we just change the name to scripts?

Done

PR #85, opened 2018-04-15
Loader improvements

cdce8p:
Improvements to the loader class:
 - Added `char.from_dict` and `service.from_dict` methods
 - Loader refactoring
 - Removed `char_class`, `service_class`. I don't think those are necessary.


pyhap/service.py
```diff
@@ -68,3 +71,15 @@ def to_HAP(self):
...
+
+    @classmethod
+    def from_dict(cls, name, json_dict, char_loader):
+        """Initialize a service object from a dict."""
+        type_id = UUID(json_dict.pop('UUID'))
+        service = cls(type_id, name)
+        chars = []
+
+        for name in json_dict['RequiredCharacteristics']:
+            chars.append(char_loader.get(name))
```
ikalchev: Is there a reason to prefer this to
```python
for name in json_dict[...]:
    service.add_characteristic(char_loader.get(name))
```
Personally I think this is more readable

ikalchev:
Much simpler, kudos! Just a few in-line comments

pyhap/characteristic.py
```diff
@@ -216,3 +218,9 @@ def to_HAP(self):
             hap_rep['value'] = self.value
 
         return hap_rep
+
+    @classmethod
+    def from_dict(cls, name, json_dict):
```
ikalchev: Should we add docs for the required fields?

pyhap/service.py
cdce8p: I agree. Must have been so obvious that I've overlooked it.

pyhap/characteristic.py
cdce8p: If you like me to, I will do them.

However personally I'm not a big fan of doc comments for obvious things. I don't remember where I've heard it, but someone once said: 'If you need comments for your code, then the code isn't good enough.'
In a broader context, I think the name of a function or method and its parameter should be enough to understand their meaning. For any method that is a direct way to interact with `HAP-python` I might understand it, but I believe it isn't necessary here. I would even go as far as to say that most of the comments for `TypeLoader`, `CharLoader` and `ServiceLoader` are unnecessary. Who ever needs to call these explicitly, probably can read python.

cdce8p:
I did some last changes:
 - The `CharLoader` and `ServiceLoader` have just been overriding the `get` method. I combined them in the `TypeLoader` with the methods `get_char` and `get_service`. Added additional checks that the `char_dict` / `service_dict` contain the necessary keys.
 - I extracted the unrelated changes from #86 (permission change and `update_advertisement`) to include them here.

pyhap/characteristic.py
ikalchev: True, but comments are used throughout and consistency always helps. My point was that a simple list with bullets like
- uuid
- HAP properties
    - format
is much less ambiguous and easy than to read than code. Anyway, I don't mind either way.

pyhap/loader.py
```diff
@@ -33,145 +35,67 @@ def __init__(self, fp):
...
+        :raise KeyError: When characteristic file did not contain necessary
+            keys.
         """
-        super(CharLoader, self).__init__(fp)
-        self.char_class = char_class
-
-    def get(self, name, char_class=None):
-        """Instantiate and return a `char_class` with the given name.
+        char_dict = self.get(name)
+        if 'Format' not in char_dict or 'Permissions' not in char_dict or \
```
ikalchev: Perhaps split  each check on new line

pyhap/loader.py
```diff
@@ -33,145 +35,67 @@ def __init__(self, fp):
...
-        serv_info = super(ServiceLoader, self).get(name)
-        type_id = uuid.UUID(serv_info["UUID"])
-        service_type = service_class or self.service_class
-        s = service_type(type_id, name)
-        chars = [self.char_loader.get(c) for c in serv_info["RequiredCharacteristics"]]
-        s.add_characteristic(*chars)
-        return s
+        char_loader = char_loader or get_char_loader()
+        service_dict = self.get(name)
+        if 'RequiredCharacteristics' not in service_dict or \
```
ikalchev: Same as before

PR #87, opened 2018-04-16
Travis - Removed py34

cdce8p:
Async syntax is first supported in py3.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ branches:
     - /^\d\.\d+\.\d+(rc\d+|\.dev\d+)?$/
 matrix:
   include:
-    - python: 3.4
-      env: TOXENV=py34
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6


ikalchev:
Sorry for that I only removed 3.4 from tox

ikalchev:
This is strange, travis builds are failing sometimes, probably haven't mocked the aio/thread stuff. Need to check this.

anyway, thanks for this.

cdce8p:
Since it fails for the `accessory_driver`, it might be related to the change to `run_forever`. If that's ok with you, I would like to redo / fix the tests in a separate PR. That way it might be a bit easier for me and I can include #83, #84, #85 and #86 in one. It would be great if #85 and #86 are merged of this.

I hope to get this done by the end of the day.

ikalchev:
It’s fine, it could be that I haven’t patched the tests properly. I expect to have limited availability today and tomorrow, but will try to review/merge if something comes along

PR #89, opened 2018-04-16
Fixed accessories

cdce8p:
This PR should fix the accessory files. It requires #88 to fix them completely.


ikalchev:
We need to compile a big fat list of breaking changes for the next release. I will try to do this soon as it is important

Issue #78, opened 2018-04-04
[Feature Request] getter callback


thomaspurchas:
I noticed that in all the example accessories there is always some sort of polling loop updating a char `value`.

I think it would be good to add a `getter_callback` (in addition to the `setter_callback`) that is called when an iOS client requests a char `value`. Then the request can trigger a sensor poll, returning the most recent data, updating the local `value`, and sending any notifications.

This would also remove the need for an explicit polling loop in most cases.

ikalchev:
Sounds good. FYI iOS subscribes for almost all characteristics, so it waits for changes and expects these to be updated without it having to ask for it with a GET request. The only time when iOS does a GET is when you open your Home app or pull your control center and tap on the home icon thingy. This was the reason I decided to go for a more push-y model.

But I def agree - a getter callback can make things simpler for same cases and allows for more choice

PR #90, opened 2018-04-17
Added getter_callback to Characteristic

jslay88:
For #78


cdce8p:
@jslay88 Can you rebase your branch into to current dev branch? That should solve the merge conflicts.

cdce8p:
As far as I understand it, the `getter_callback` should be used to pull updates from a sensor within `client_update_value`. Wouldn't it then be enough to make value optional and add
```py
if self.getter_callback:
    value = self.getter_callback()
```

jslay88:
Yeah, you are probably right. This is what I get for trying to put something in late at night waiting on things to finish. Give me a day, I want to rework this now that I am looking at it again.

jslay88:
Should we change the setter_callback to not use value as well? Seeing as how it is implied through `char.value` property anyway?

pyhap/characteristic.py
```diff
@@ -98,14 +98,39 @@ def __init__(self, display_name, type_id, properties):
...
+
+        To set the value for the `Characteristic` on a state
+        change, use `set_value` to handle notifying clients.
+
+        .. seealso:: Characteristic.set_value()
+
+        :return: value
+        """
+        if not getattr(self, '_value', None):
+            self._value = self._get_default_value()
```
cdce8p: Why not leave that in `__init__` with `self._value = self.get_default_value()`? Currently you do the check every time `char.value` is called, also it's only `None` for the first time.

pyhap/characteristic.py
```diff
@@ -98,14 +98,39 @@ def __init__(self, display_name, type_id, properties):
...
+        if self.getter_callback:
+            self._value = self.to_valid_value(self.getter_callback())
```
cdce8p: Maybe change that to `self._value = self.to_valid_value(value=self.getter_callback())`.
By adding `value=` it might be a bit clearer what `self.getter_callback()` will return.

pyhap/characteristic.py
```diff
@@ -158,16 +183,22 @@ def override_properties(self, properties=None, valid_values=None):
             self.properties[PROP_VALID_VALUES] = valid_values
 
         try:
-            self.value = self.to_valid_value(self.value)
+            self.set_value(self.to_valid_value(self.value), should_notify=False)
         except ValueError:
-            self.value = self._get_default_value()
+            self.set_value(self._get_default_value(), should_notify=False)
```
cdce8p: `override_properties` shouldn't call `set_value`. It would only trigger unnecessary calls, like the log output.
Just use:
```py
        try:
            self._value = self.to_valid_value(self.value)
        except ValueError:
            self._value = self._get_default_value()
```

pyhap/characteristic.py
```diff
@@ -177,19 +208,20 @@ def set_value(self, value, should_notify=True):
...
+        Call set_value and call callback.
         """
         logger.debug('%s: Client update value to %s',
-                      self.display_name, value)
-        self.value = value
-        self.notify()
+                     self.display_name, value)
+        # Set new value before anything else, set_value calls notify
+        self.set_value(value)
+        # Call setter_callback
```
cdce8p: The same here. It would be quite confusion if a `client_update_value` call cause a `set_value` log output as well.
Use `self._value = value` instead.

tests/test_characteristic.py
```diff
@@ -172,14 +172,14 @@ def test_to_HAP_numberic(self):
...
+        char.set_value('aaa')
         with patch.object(char, 'broker') as mock_broker:
             hap_repr = char.to_HAP()
         self.assertEqual(hap_repr['format'], 'string')
         self.assertNotIn('maxLen', hap_repr)
 
-        char.value = 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee' \
-            'ffffffffffgggggggggg'
+        char.set_value('aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee'
+                       'ffffffffffgggggggggg')
```
cdce8p: Personally I would prefer `char._value = ` instead. That would make it a bit clearer what is being tested and you don't have to be that careful what is being mocked when.

ikalchev:
I think a more straightforward solution would be to make a new interface for getting a value from iOS clients. The calls that get the char value to give to iOS is in `AccessoryDriver.get_characteristics` - currently it does only `char.value`. We can define a new method for Characteristic that gets the iOS value. Actually this was the idea of the recently-removed `get_value`. We can get it back and use that to do
```python
def get_value(self):
    if self.getter_callback:
        return self.getter_callback()
    else:
        return value
```
However, we also need to use this when getting the value in the `to_HAP`.

What are your thoughts?

PS sorry for the late review!

jslay88:
Well, that's kind of the intent behind making `value` a property. That way, it is intuitive for `get` just by accessing `char.value`. Maybe I am over/under thinking it though. Using `char.value` we wouldn't have to go and update a bunch of code outside of `characteristic.py`

cdce8p:
The breaking change with using the `property` decorator wouldn't be that big of a problem, since the user interacts with `set_value` and not `value`.
I didn't though of the time aspect though and that makes sense.

What I don't quite get is, why `get_value` should be used in `to_HAP`. The way I image it at the moment is the following:
 - The driver methods to get the value of a char calls `get_value`
 - `get_value` updates the value internally for the char, it can return the new value
 - The driver checks if the value has changed. If so it either retrieves the new or uses the returned one to send an update to HomeKit.
 - In both cases `char.value` is equal to the current value.
 - `to_HAP` isn't called.

An additional argument for not including the `get_value` call in `to_HAP` is that separating those two things makes the conversion to async easier later.

ikalchev:
I didn’t quite get point 3 - the driver does not check the value - it just returns it. 

My argument about using get value in to HAP is that othwewise get_accessories and get_characteristics will have different semantics (one uses to hap, the other .value)

cdce8p:
> I didn’t quite get point 3 - the driver does not check the value - it just returns it.

I could have been a bit clearer. My idea for `get_value` was to implement a method on the service level to update the value regularly, something like the `@Accessory.run_at_interval` method at the moment. In that case the driver would only need to send and update to HomeKit if the value has changed between the last two calls.

> My argument about using get value in to HAP is that othwewise get_accessories and get_characteristics will have different semantics (one uses to hap, the other .value)

I see. My idea was separation of `get_value` and `to_HAP`. While I'm thinking about it, do we really need to call `get_value` in either of those methods? For `get_accessories` sending the default values should be enough and `get_characteristics` could just get the current char value (without updating it before). That would play quite nicely with my idea above, to call `get_value` regularly for the accessory and then to push changes.

ikalchev:
Basically I see two use cases for how the value is retrieved - you retrieve a cached value or get it anew, where the latter is achieved with setting the getter_callback. Your point about regularly refreshing the value falls into the first case and can be achieved using the already established approach with the `Accessory.run` method.

Having said that, I think it is flexible enough to have a `get_value` method (because the property does not make it obvious that there could be delays) and an optional `getter_callback` which is called within it. In this way, we leave it to the user to choose the drawback most suitable for her -(1) increased latency in `to_HAP` because of the `getter_callback`, but minimum interaction with the "device" or (2) fast `to_HAP`, because of no `getter_callback`, but increased overhead of regularly updating the value.

From the set of sensors I have, there are some which have limited lifetime and these should be "read from" as rarely as possible. Thus, I have a 20/30 min `run_at_interval` for them currently and would pretty much like a `getter_callback`. However, for others I have a 1/2 min interval which works fine as it is.

Thoughts?

cdce8p:
Please correct me if I haven't understand something properly.

What it comes done to, as I think of it, is to use `get_value` only inside the `Accessory.run` method. Basically adding a helper method to make calling a function to retrieve a value easier. I don't see any real value in a callback in another situation.

In that case it might make sense to expand upon the current idea of using just:
```py
def get_value():
    if self.getter_callback:
        self._value = self.to_valid_value(value=self.getter_callback())
    return self._value
```
and incorporate the notify call if the new value is different. Something like:
```py
def get_value():
    if not self.getter_callback:
        raise ValueError('No getter_callback defined')
    value = self.to_valid_value(value=self.getter_callback())
    if value != self.value:
        self.value = value
        self.notify()
```
Maybe a renaming to `getter_helper` or something else would work as well.

jslay88:
What if we were to have a bool on the char to define the state's requirements. 
There are some cases where a char should use the stored state, and there are some where it should ALWAYS pull a new state, even if it is the same state as stored. Then, we can have, instead of getter_callback, update_value_callback, or something else. 

Just throwing more ideas out there. I really have no weight in which method we go with.

ikalchev:
@cdce8p I think getting a value should not trigger a notify, because the "gotten" value will be deliver to iOS anyway. Getter callback is intuitive for me - method to be called when a get happens.

@jslay88 I think the getter should decide on that internally, i.e. no need to expose the bool in the char.

Can we agree on:
```python
def get_value():
    if self.getter_callback:
        self.value = self.to_valid_value(value=self.getter_callback())
    return self.value
```
with no `_value` and calling this in to_HAP and the driver methods? Do we have any pain points and disagreements left?

cdce8p:
We can do that, although small changes for the async conversion might be necessary later.

ikalchev:
@jslay88 Are you fine with the above?

jslay88:
Sure thing. Will make the changes here shortly.

pyhap/characteristic.py
```diff
@@ -184,12 +196,15 @@ def set_value(self, value, should_notify=True):
     def client_update_value(self, value):
         """Called from broker for value change in Home app.
 
-        Change self.value to value and call callback.
+        Call set_value and call callback.
         """
-        logger.debug('client_update_value: %s to %s',
-                      self.display_name, value)
+        logger.debug('%s: Client update value to %s',
```
cdce8p: I changed the log statement in one of the last merged PR #99

pyhap/characteristic.py
```diff
@@ -184,12 +196,15 @@ def set_value(self, value, should_notify=True):
...
+                     self.display_name, value)
+        # Set new value before anything else, then notify client
+
```
cdce8p: This and the comment below shouldn't be necessary.

tests/test_characteristic.py
```diff
@@ -202,3 +202,162 @@ def test_from_dict():
     assert char.display_name == 'Test Char'
     assert char.type_id == uuid
     assert char.properties == {'Format': 'int', 'Permissions': 'read'}
+
+class TestCharacteristic(unittest.TestCase):
```
cdce8p: All tests in the `TestCharacteristic` class got changed in #99 as well. They are now outside, so you should be able to remove the class.

pyhap/characteristic.py
```diff
@@ -114,6 +115,11 @@ def _get_default_value(self):
             value = HAP_FORMAT_DEFAULTS[self.properties[PROP_FORMAT]]
             return self.to_valid_value(value)
 
+    def get_value(self):
```
ikalchev: Documentation stating that this calls the getter if any to update the .value

ikalchev:
@jslay88 any issues with this?

jslay88:
I blasted away at it, I think I got everything requested between @ikalchev and @cdce8p.

Please double check, as I was quickly rushing through this.

pyhap/characteristic.py
```diff
@@ -221,10 +236,10 @@ def to_HAP(self):
             hap_rep.update({k: self.properties[k] for k in
                             self.properties.keys() & PROP_NUMERIC})
         elif self.properties[PROP_FORMAT] == HAP_FORMAT_STRING:
-            if len(self.value) > 64:
-                hap_rep[HAP_REPR_MAX_LEN] = min(len(self.value), 256)
+            if len(self.get_value()) > 64:
```
ikalchev: Maybe do `value = self.get_value()` at the beginning of the method, otherwise you could potentially call the callback two times in this if

jslay88: Yup, thanks for checking. Pushed changes.

pyhap/characteristic.py
```diff
@@ -184,12 +200,11 @@ def set_value(self, value, should_notify=True):
...
         """Called from broker for value change in Home app.
 
-        Change self.value to value and call callback.
+        Call set_value and call callback.
         """
-        logger.debug('client_update_value: %s to %s',
-                      self.display_name, value)
         self.value = value
         self.notify()
+        # Call setter_callback
```
cdce8p: Can you please revert the changes to `client_update_value`? I don't think they are necessary for `get_value`.

pyhap/characteristic.py
```diff
@@ -217,14 +232,16 @@ def to_HAP(self):
             HAP_REPR_FORMAT: self.properties[PROP_FORMAT],
         }
 
+        value = self.get_value()
+
```
cdce8p: A small one: `value` is used in the block below, so why not delete the empty line?

jslay88:
So do you want the logger in `client_update_value` or not? You requested to have it removed, and now are wanting it back?


`This and the comment below shouldn't be necessary.`

cdce8p:
I see. The comment earlier was meant for `# Set new value before anything else, then notify client` and `# Call setter_callback`, not the logger statement. I should have been more precise, sorry about that.

jslay88:
Done

cdce8p:
Thanks! The code looks good IMO.
If you could do the last finishing touch and add this change to `CHANGELOG.md` that would be really great. You might need to rebase to avoid merge conflicts, before you do that.

jslay88:
Done

ikalchev:
Great work, thanks for addressing all comments!

PR #92, opened 2018-04-20
Improvement suggestions

cdce8p:
A collection of improvement suggestions. I apologize beforehand, this is quite much. To make reviewing it a bit easier, I did my best splitting the individual changes into separate commits.

The central ideas:
 - I have learned recently that `untitest.TestCase` tests are slower then regular `asserts`. Therefore I converted `test_characteristic`. #99
 - To improve readability in the accessory module, `IIDManager` now has it's own module #100
 - Small refactoring for `IIDManager` to improve performance slightly (only one dict instead of two) #100
 - Added tests for `IIDManager` class #100
 - Rearranged requirements (files), sorted alphabetically #101
 - Renaming: `display_name` to just `name` (breaking-change)
 - Renaming: `broker` for accessory to `driver` #104
 - Changed `""` to `''` for variable names, more pythonic #104
 - Updated accessories to account for changes

Bigger changes:
 - Removed the `_set_services` method and replaced it with `add_info_service` / `set_info_service` for the `AccessoryInformation` service. That way we make sure it is always `iid=1`. Alternative to #41  (breaking-change) #102
 - To make adding services a bit easier, I added the helper method `add_preload_service(service, chars)` which takes in the name of a service and a string / list of char names and auto-loads the service. #102
 - Make QR code optional to use. That way we can remove to dependencies. If the user wants to, he can enable it quite easily after he installed the deps. #103

Let me know if you have any questions or suggestions.

### DIFF
--- a/main.py
+++ b/main.py
@@ -20,7 +20,7 @@ logging.basicConfig(level=logging.INFO)
 
 def get_bridge():
     """Call this method to get a Bridge instead of a standalone accessory."""
-    bridge = Bridge(display_name='Bridge')
+    bridge = Bridge(name='Bridge')
     temp_sensor = TemperatureSensor('Sensor 2')
     temp_sensor2 = TemperatureSensor('Sensor 1')
     bridge.add_accessory(temp_sensor)
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -28,16 +28,16 @@ class Accessory:
     category = CATEGORY_OTHER
 
     @classmethod
-    def create(cls, display_name, pincode, aid=STANDALONE_AID):
+    def create(cls, name, pincode, aid=STANDALONE_AID):
         mac = util.generate_mac()
-        return cls(display_name, aid=aid, mac=mac, pincode=pincode)
+        return cls(name, aid=aid, mac=mac, pincode=pincode)
 
-    def __init__(self, display_name, aid=None, mac=None, pincode=None,
+    def __init__(self, name, aid=None, mac=None, pincode=None,
                  iid_manager=None, setup_id=None):
         """Initialise with the given properties.
 
-        :param display_name: Name to be displayed in the Home app.
-        :type display_name: str
+        :param name: Name to be displayed in the Home app.
+        :type name: str
 
         :param aid: The accessory ID, uniquely identifying this accessory.
             `Accessories` that advertised on the network must have the
@@ -61,7 +61,7 @@ class Accessory:
             4 digit string 0-9 A-Z
         :type setup_id: str
         """
-        self.display_name = display_name
+        self.name = name
         self.aid = aid
         self.mac = mac
         self.config_version = 2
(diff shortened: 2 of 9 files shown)


ikalchev:
I won’t have time to go through all the changes now, but some points:

- test case - why are you worried about the performance of the tests?
- display_name is a perfectly fine name - changing it just introduces a breaking change. I don’t see much value it in.
- Make QR usage the default please. I know you want to decrease HA dependencies and I want to help with that, but I think HAP python should be comfortably usable as a stand-alone package. So users should get WR by default and disabling it should be optional.
- I need to go through the add preload services to comment.

I hope I can review this over the weekend.

cdce8p:
> test case - why are you worried about the performance of the tests?

Not really. At this scale it shouldn't matter. However, since I did the tests for `loader`, `service` and `iid_manager` without `unittest.TestCase` and `test_characteristic` was using both, I wanted to adapt it.

> display_name is a perfectly fine name - changing it just introduces a breaking change. I don’t see much value it in.

Technically it is a breaking change, although a very small one that doesn't effect everybody. Only if the call for the accessory was with the keyword. For the `accessories` that was only once in `main` the case. The value I see is for new users. I always got confused when I started using pyhap that it is `display_name` instead of `name`. If you still want to, I will remove that commit and revert the change.

> Make QR usage the default please. I know you want to decrease HA dependencies and I want to help with that, but I think HAP python should be comfortably usable as a stand-alone package. So users should get QR by default and disabling it should be optional.

My initial approach might have been a bit drastic. I took that from the way `pyqrcode` handles it for [png output](https://pythonhosted.org/PyQRCode/moddoc.html#pyqrcode.QRCode.png). I though about how this could be improved and came up with two options:
 1. Install requirements during first run if `qr_code=True`
 2. Define an additional `extras_require`. That way the QR code install would just be `pip3 install HAP-python[QRCode]`

Personally I would go with `2.`. This would make overriding `setup_message` unnecessary as well. It would be a smaller price to pay for removing two dependencies.

cdce8p:
I added `extras_require` for the QRCode and fixed a bug in my new service methods.
Leave a note if you start looking at the code, just so that don't accidentally rebase something you've already looked it.

cdce8p:
@ikalchev Did you had the chance to take a look at some of this PR already?
If it makes reviewing easier, I can split it up into several smaller once.

cdce8p:
No problem. It would just require a bit coordination on our side when merging those, since they depend on each other and therefore should preferable be merged in a specific order.

cdce8p:
I started with extracting the first commits. If you approve of the new PR's, they can be merged in order hopefully without merge conflicts. Otherwise I will rebase them.

Unfortunately I can't do the others until tomorrow (or even Wednesday), but those three should be enough for the beginning.

ikalchev:
I just merged them :) Take your time, I will try to focus on the camera now, I see there are a few requests for this.

cdce8p:
I've done another set/split of three PR's to review. For those, as long as there aren't any merge conflicts, the order shouldn't matter.

cdce8p:
Closing this QR.
Changes have been extracted to separate PR's.

Still open:
 - QR-Code, see #103 
 - Change to `name`. Dropping it, since it might not be that necessary.

PR #93, opened 2018-04-21
Fix aio_stop_event

cdce8p:
In addition to #83, `asyncio.Event()` needs to be called with the `loop` parameter. Otherwise it throws an error when not executed in the main thread.


cdce8p:
It even fixed one of the two failing tests. Unintentionally 😅

cdce8p:
I fixed the last failing test as well. It never reached `driver.stop()`, since it never exited `driver.start()`.

cdce8p:
Can this PR be merged (after #96 is merged)? It would make developing for me a bit easier.

ikalchev:
Yes sorry give me a few hours and I will merge both

cdce8p:
Thanks :+1:

PR #102, opened 2018-05-01
Improvements 4 - Service Helper

cdce8p:
Split of from #92.

### Changes
- Added helper method to enable easy override of the `AccessoryInformation` service.
- Added helper method to load a service and chars and add it to an accessory.
- The `AccessoryInformation` service will always have the `iid=1`.
- Updated accessories to account for changes

### Breaking Changes
- Changed default values associated with the `AccessoryInformation` service.
- Removed accessory call to `Accessory._set_services`. Overriding this method not longer works. Instead services should be initialized in the accessories `init` method.

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -85,6 +85,8 @@ class Accessory:
 
         self.add_info_service()
 
+        self._set_services()
+
     def __repr__(self):
         """Return the representation of the accessory."""
         services = [s.display_name for s in self.services]
@@ -108,6 +110,16 @@ class Accessory:
         if self._pincode is None:
             self._pincode = util.generate_pincode()
         return self._pincode
+
+    def _set_services(self):
+        """Sets the services for this accessory.
+
+        This method is now deprecated, initialize the service inside the
+        accessory `init` method.
+        """
+        logger.warning(
+            "The 'Accessory._set_services' method is deprecated. Initialize "
+            "the services inside the accessories 'init' method instead.")
 
     def add_info_service(self):
         """Helper method to add the required `AccessoryInformation` service.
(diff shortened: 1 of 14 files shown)


pyhap/accessory.py
```diff
@@ -109,27 +109,46 @@ def pincode(self):
...
+        if serial_number:
+            if len(serial_number) >= 1:
+                serv_info.configure_char('SerialNumber', value=serial_number)
+            else:
+                logger.warning(
+                    "Couldn't add SerialNumber for %s. The SerialNumber must "
+                    "be at least one character long.", self.display_name)
+
+    def add_preload_service(self, service, chars=None):
+        """Define and return a service to be available for the accessory."""
```
ikalchev: `Create a service with the given name and add it to this accessory`?

pyhap/accessory.py
```diff
@@ -109,27 +109,46 @@ def pincode(self):
             self._pincode = util.generate_pincode()
         return self._pincode
 
-    def _set_services(self):
```
ikalchev: I would really like backwards compatibility with _set_services. Do you think we can change the base implementation of it to not add an info service, so old clients can still use it? This way we will avoid a breaking change.

pyhap/accessory.py
```diff
@@ -109,27 +109,46 @@ def pincode(self):
...
+            else:
+                logger.warning(
+                    "Couldn't add SerialNumber for %s. The SerialNumber must "
+                    "be at least one character long.", self.display_name)
+
+    def add_preload_service(self, service, chars=None):
+        """Define and return a service to be available for the accessory."""
+        service = get_serv_loader().get_service(service)
+        if chars:
+            chars = chars if isinstance(chars, list) else [chars]
```
ikalchev: Let's expect a single type here (a list)

cdce8p: I would like to leave it in there and except a single char as string, since there are actually quite a lot of situations where you want to add **only one** other char.

pyhap/accessory.py
cdce8p: I removed the breaking change and marked the `_set_service` method as `deprecated` instead. That way it should still work without issues.

ikalchev: Have you tested? If not, I can do the testing for this and any necessary changes after this merge.

ikalchev: Okay, then I am merging this and will try to test the backwards compatibility.

PR #103, opened 2018-05-01
Improvements 5 - QR Code

cdce8p:
Split of from #92.

### Changes
- `pip install HAP-python[QRCode]` now installs all dependencies inclusive those to display the QR-Code during setup.

### Breaking Changes
- QR Code is now optional

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -4,9 +4,7 @@ import logging
 import struct
 import threading
 
-import base36
 import ed25519
-from pyqrcode import QRCode
 
 from pyhap import util
 from pyhap.const import (
@@ -255,12 +253,12 @@ class Accessory:
     def paired(self):
         return len(self.paired_clients) > 0
 
-    @property
     def xhm_uri(self):
         """Generates the X-HM:// uri (Setup Code URI)
 
         :rtype: str
         """
+        import base36
         buffer = bytearray(b'\x00\x00\x00\x00\x00\x00\x00\x00')
 
         value_low = int(self.pincode.replace(b'-', b''), 10)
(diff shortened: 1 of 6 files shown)


ikalchev:
As I previously said - I would like HAP-python to have features such as the QR code out of the box. I understand you want to reduce the dependencies because of HA, but is there way for the QR code to be enabled by default and users can do something like e`pip install HAP-python[no-QRCode]`?

cdce8p:
Unfortunately that isn't supported, at least as far as I found. The intended use case is an opt in to additional features, not an opt out. At least in my option this makes sense, since only the dependencies that are required for the core functions are installed that way.

http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

I've changed the install note in the `readme`. Sadly that's the best I could do.

pyhap/accessory.py
```diff
@@ -4,9 +4,7 @@
 import struct
 import threading
 
-import base36
 import ed25519
-from pyqrcode import QRCode
```
ikalchev: Small request, if it's not too much: Can you do it like a feature switch:
```python
try:
   from pyqrcode import QRCode
   HAS_QR = True
except ImportError:
   HAS_QR = False
...
...
if not HAS_QR:
   print("Enter the below code...")
print(The QR code)
```
I think this will make the code easier to read and maintain, because you still keep the imports in one place and the logic in another.

As an optimisation, I think this can even be done in the __init__py:
```python
#__init__.py

try:
   import pyqrcode
   HAS_QR = True
except ImportError:
   HAS_QR = False

try:
   import X
   HAS_X = True
except ImportError:
   HAS_X = False

# foo.py
from pyhap import HAS_QR
if HAS_QR:
   import pyqrcode

# bar.py
from pyhap import HAS_X
if HAS_X:
   import X
```

pyhap/accessory.py
```diff
@@ -294,10 +279,23 @@ def to_HAP(self):
...
+            from pyqrcode import QRCode
+            xhm_uri = self.xhm_uri()
+            print('Setup payload: {}'.format(xhm_uri), flush=True)
+            print('Scan this code with your HomeKit app on your iOS device:',
+                  flush=True)
+            print(QRCode(xhm_uri).terminal(quiet_zone=2), flush=True)
+            print('Or enter this code in your HomeKit app on your iOS device: '
+                  '{}'.format(self.pincode.decode()))
+        except ImportError:
+            print('Enter this code in your HomeKit app on your iOS device: {}'
```
ikalchev: Maybe add: To use the QR code feature, pip install HAP-python[QRCode]

PR #104, opened 2018-05-01
Improvements 6 - Changes, Driver

cdce8p:
Split of from #92.

### Changes
- Replaced double `"` for parameter values with `'`.
- Removed `object` inheritance
- Removed class from super calls if obvious

### Breaking Changes
- Changed `Accessory.broker` to `Accessory.driver`. Also `acc.set_broker` is now `acc.set_driver`.

---

Also the change to `driver` might not be entirely necessary, it simplifies things with the better name.


ikalchev:
The set_broker is unlikely to be used a lot by users, so it's not too big a deal that this is a breaking change.

PR #105, opened 2018-05-02
Improvements 6 - Driver assignment

cdce8p:
During my improvement PR's I noticed that we assign may things to each accessory although they are only needed once (at the driver level). Especially for bridged accessories that is mostly unnecessary. This PR should be a first step in reversing this behavior.

Instead of assigning the accessory to the driver, I think that I would make much more sense to assign the driver to each accessory and later add the top level one to the driver.

### To-Do
- [x] Remove `accessory.set_sentinel`
- [x] Use `driver.run_sentinel`, `driver.aio_stop_event` and `driver.event_loop`
- [x] ~~Maybe: Assign `keys`, `pincode`, `mac` to driver instead.~~
- [x] Save loader in the driver.

**Note**: All these changes assume that we stick to the current version, where one driver only handles one top level accessory.

@ikalchev What do you think this?

### DIFF
--- a/pyhap/loader.py
+++ b/pyhap/loader.py
@@ -8,14 +8,16 @@ instance of it (as long as it is described in some
 json file).
 """
 import json
-import logging
 
 from pyhap import CHARACTERISTICS_FILE, SERVICES_FILE
 from pyhap.characteristic import Characteristic
 from pyhap.service import Service
 
-_loader = None
-logger = logging.getLogger(__name__)
+
+def read_file(path):
+    """Read file and return a dict."""
+    with open(path, 'r') as file:
+        return json.load(file)
 
 
 class Loader:
@@ -28,13 +30,8 @@ class Loader:
     def __init__(self, path_char=CHARACTERISTICS_FILE,
                  path_service=SERVICES_FILE):
         """Initialize a new Loader instance."""
-        self.char_types = self._read_file(path_char)
-        self.serv_types = self._read_file(path_service)
-
-    def _read_file(self, path):
-        """Read file and return a dict."""
-        with open(path, 'r') as file:
-            return json.load(file)
+        self.char_types = read_file(path_char)
+        self.serv_types = read_file(path_service)
 
     def get_char(self, name):
         """Return new Characteristic object."""
@@ -60,38 +57,3 @@ class Loader:
         loader.char_types = char_dict or {}
         loader.serv_types = serv_dict or {}
         return loader
-
-
-def get_loader():
-    """Get a service and char loader.
-
-    If already initialized it returns the existing one.
-    """
-    global _loader
-    if _loader is None:
-        _loader = Loader()
-    return _loader
-
-
-# pylint: disable=unused-argument
-def get_char_loader(desc_file=None):
-    """Get a CharacteristicLoader with characteristic descriptions in the given file.
-
-    .. deprecated:: 2.0
-       Use `get_loader` instead.
-    """
-    logger.warning(
-        "'get_char_loader' is deprecated. Use 'get_loader' instead.")
-    return get_loader()
-
-
-# pylint: disable=unused-argument
-def get_serv_loader(desc_file=None):
-    """Get a ServiceLoader with service descriptions in the given file.
-
-    .. deprecated:: 2.0
-       Use `get_loader` instead.
-    """
-    logger.warning(
-        "'get_serv_loader' is deprecated. Use 'get_loader' instead.")
-    return get_loader()
--- a/pyhap/loader.py
+++ b/pyhap/loader.py
@@ -8,16 +8,14 @@ instance of it (as long as it is described in some
 json file).
 """
 import json
+import logging
 
 from pyhap import CHARACTERISTICS_FILE, SERVICES_FILE
 from pyhap.characteristic import Characteristic
 from pyhap.service import Service
 
-
-def read_file(path):
-    """Read file and return a dict."""
-    with open(path, 'r') as file:
-        return json.load(file)
+_loader = None
+logger = logging.getLogger(__name__)
 
 
 class Loader:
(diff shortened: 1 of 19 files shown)


ikalchev:
Few points:
- I think that an Accessory should be able to exist without a driver
- The load and persist methods of the driver allow users to overwrite as necessary. There were comments at one time when someone had a read-only file system on the raspberry - one option to make hap work on this is to overwrite the persist/load to do that to/from a remote store.
- Keeping the "management" properties in the driver sounds good( and obvious, now that I think about it)!

cdce8p:
> The load and persist methods of the driver allow users to overwrite as necessary.

I reverted the changes to the encoder and persistence methods.

> I think that an Accessory should be able to exist without a driver

Is there a reason for it? As far as I see it, accessories can't do anything useful without a driver. That is even less if you consider that the driver handles the event loop.

> Keeping the "management" properties in the driver sounds good( and obvious, now that I think about it)!

Maybe we should outsouce them to a entirely new class, similar to the `IID_Manager`? However this shouldn't be included in this PR.

ikalchev:
> Maybe we should outsouce them to a entirely new class?
Yes, maybe AccessoryState or something.

> Is there a reason for it?
No reason, but why couple them?

cdce8p:
> No reason, but why couple them?

They are already. What reason is there for an accessory to exist without the driver. Currently it's just the other way round which complicates thing IMO. Passing the driver to each `accessory` allows us to shortcut some functions.

---

I will update this PR. That should make it clearer where the benefits are.

ikalchev:
What do you mean by "complicates things"? The way I see it is that there are three clear, separate things: (1) A representation of a service (2) the state of the service and (3) a thing which manipulates the state of the service as a result of interactions with the outside world.

Currently, the Accessory is both the representation of a service and its state and the driver is the thing that manipulates. What this change proposes is to move the state from the accessory to the driver, which is totally fine, although ideally we should separate it. However, by forcing the accessory to have a driver, we essentially need to have all three things created to get functionality even from only one of them. For example, to test the Accessory I have to have a driver. I don't see this is as bringing any benefits. Right now I can have and Accessory and a Driver separately, potentially reusing and substituting whatever I want individually.

cdce8p:
Let me step back a few points.

The problem I see with the current setup is that we create the event loop within the driver (which is the right call IMO). As we move more and more methods to async. Eventually nearly all objects will need a reference to the `loop` to work.

To be honest I took this lesson from Home Assistant as well. Their we have the `hass` object as a starting point (and the event loop handler) and basically any object created has some kind of reference to it.

----

> separate things

Their is nothing against that. The question that remains however is how much sense does it make to have a object that is totally separated, but can't do anything since it requires some setup (eg. `set_driver` and `set_sentinel` for the accessory)?

> Tests

You're right their. However currently you need an accessory to test the top level driver. HA has a special function to handle this `get_test_home_assistant()`. It returns a `hass` object to use for tests.

> Move state from accessory to driver

I still plan on doing this.

---

The way I see it the hierarchy for `HAP-python` looks / could look something like this:

```text
-> AccessoryDriver
  -> AccessoryState (or a different name)
  -> Loader (should be cached here instead of the module)
  -> Encoder
  -> HAPServer
    -> ...
  -> Accessory
    -> IIDManager
    -> Service
      -> Characteristic
```

ikalchev:
Will have another look when you resolve the conflicts, but I think we are mostly there.

cdce8p:
@ikalchev I have updated this PR.
For a complete list of changes see: [Changelog/unrelease](https://github.com/cdce8p/HAP-python/blob/9f8c187514f84cc9abad7f2c41e63962d44c2064/CHANGELOG.md#unreleased)

The main ideas some of which we already discussed:
 1. Assign `driver` to accessory during `init`
 2. Add `acc` to driver with `driver.add_accessory()`
 3. `acc.set_driver()` and `acc.set_sentinel()` were redundant and have therefore been removed.
 4. `run_sentinel`(now `stop_event`), `aio_stop_event` and `loop` are now accessible through the `driver`.

Additional changes:
 **5.** Added default `port` to `accessory_driver`: `51234`. That's the one you used in the tests.
 **6.** The `loader` object is now stored inside the `driver`, instead of in the module. Hence `get_loader()`, `get_serv_loader()` and `get_char_loader()` can be removed entirely.
 **7.** `driver.init()` now requires keyword arguments. That will make any future changes easier and we already have a breaking change for this method.
 **8.** Updated `accessories` and `tests`.

pyhap/loader.py
```diff
@@ -8,14 +8,16 @@
...
-import logging
 
 from pyhap import CHARACTERISTICS_FILE, SERVICES_FILE
 from pyhap.characteristic import Characteristic
 from pyhap.service import Service
 
-_loader = None
-logger = logging.getLogger(__name__)
+
+def read_file(path):
```
ikalchev: Actually `json_file_to_dict`

tests/test_accessory.py
```diff
@@ -1,58 +1,71 @@
...
+
+
+# #### Accessory ####
+# ###################
+
+def test_init():
+    Accessory(DRIVER, 'Test Accessory')
 
 
-class TestAccessory:
```
ikalchev: Why are we removing the classes? This prevents you from running tests only from a single test class. Is the performance issue really big - there only a handful of methods?

cdce8p:
> Loader changes

I've reverted them, so this PR can get merged. The original idea was just to reduce maintenance costs and avoid storing an object in a module, since this is kind of a bad practice for python.

> `json_file_to_dict`

I changed it back as well. Only added `@staticmethod` to fix a `pylint` issue. Since this is a private method and the class is small, I don't think the method name is that important.

> TestAccessory

You couldn't run them separately before either. Since `TestBridge` inherited from `TestAccessory` running this test module resulted in **9** executed tests, although it only had **7**. Additionally the change to normal methods allows us to use `pytest.fixtures` more easily (like `mock_driver`).
Furthermore I added keywords to each test. That way you know can run them individually with `pytest tests/test_accessory.py -k acc` or `pytest tests/test_accessory.py -k bridge`.

---

The `conftest.py` file I've added is for `pytest`. It is used to store fixtures that should be accessible for every test. The `MockDriver` class can be extended in the future if necessary.

PR #106, opened 2018-05-02
Test backwards compatibility of Accessory._set_services

ikalchev:


tests/test_accessory.py
```diff
@@ -9,39 +9,50 @@
...
         char.set_value(25, should_notify=True)
 
+    def test_set_services_compatible(self):
+        """Test that _set_services still works and has access to the info services"""
+        class A(accessory.Accessory):
+            def _set_services(self):
+                super()._set_services()
+                s = loader.get_serv_loader().get_service("TemperatureSensor")
+                self.add_service(s)
+                assert(self.get_service("AccessoryInformation") is not None)
```
cdce8p: `assert self.get_service("AccessoryInformation") is not None` should work

tests/test_accessory.py
```diff
@@ -9,39 +9,50 @@
...
+        a = A("Test Accessory")
+        assert(a.get_service("TemperatureSensor") is not None)
```
cdce8p: Same here

ikalchev: Thanks! Writing cpp during the day and python at night/weekend really messes up your coding habits

cdce8p:
Looks good to me :+1:

PR #108, opened 2018-05-02
Changed loader

cdce8p:
Follow up to #85
* Combine char and service loader
* Replaced `get_serv_loader`, `get_char_loader` with `get_loader`

To-Do:
- Tests
- Changelog


cdce8p:
@ikalchev Do you have the time to look over this and #107 this evening, by any chance? If the next release is out until tomorrow morning, I could update HA just in time for it's next release tomorrow evening.

ikalchev:
I will try to review this by tomorrow morning then. I understand you want this in the next HAP python release?

cdce8p:
It wouldn't be that important for HA. I just though that we could bundle the breaking changes for the loader.

What this PR comes down to basically is combining the loaders into one and thus changing to  `get_loader` instead of the other two.

pyhap/loader.py
```diff
@@ -13,90 +13,59 @@
...
 
+    @classmethod
+    def from_dict(cls, char_dict=None, serv_dict=None):
+        """Create a new instance directly from json dicts."""
+        loader = cls.__new__(Loader)
+        loader.char_types = char_dict or {}
+        loader.serv_types = serv_dict or {}
+        return loader
 
-def get_char_loader(desc_file=CHARACTERISTICS_FILE):
```
ikalchev: Let's keep this method and just forward the call to get_loader (and put a deprecated doc). Same for get_sevice_loader. I think this will avoid the breaking change

cdce8p:
Just added the deprecated methods back in.

pyhap/loader.py
```diff
@@ -8,95 +8,86 @@
...
-    if desc_file == SERVICES_FILE:
-        if _serv_loader is None:
-            with open(desc_file, 'r') as fp:
-                _serv_loader = TypeLoader(fp)
-        return _serv_loader
-
-    with open(desc_file, 'r') as fp:
-        ld = TypeLoader(fp)
-    return ld
+    logger.warning(
```
ikalchev: This caller will then fail. Can we `return get_loader()` here and for the char?

cdce8p: Yea, missed that

PR #110, opened 2018-05-04
Merge V2.0.0 into master

ikalchev:

PR #111, opened 2018-05-04
Merge master back into dev

ikalchev:


cdce8p:
The issue is, that you squash merged #110.
The best solution would be to revert the PR and merge it again with a normal merge.

cdce8p:
Although it now looks about right.

ikalchev:
Actually the problem was an additional space on a single line in the char tests

ikalchev:
I guess the space was from some PR and my editor auto-removed it (since it was trailing)

cdce8p:
In case of a normal merge, this wouldn't had been an issue for this PR which is why I suggested it. However, that works as well.

Thanks for doing the release.

ikalchev:
And thank you for the improvements 👍

PR #115, opened 2018-05-07
Remove the pyhap.accessories package.

ikalchev:
All accessories are temporarily moved to the root accessories folder.
These will be moved into separate repositories as appropriate.

pyhap.accessories is now an implicit namespace package. See
pyhap/accessories/README.md for how others can share their accessories
as HAP-python subpackages.


ikalchev:
@cdce8p @schinckel @jslay88 #43

cdce8p:
Am I right in assuming that this is the first step in extracting the `accessories` to a separate repo?

ikalchev:
Yes, and until then, I moved them to the root. Here is one example: https://github.com/ikalchev/HAP-python-AM2302

ikalchev:
Well, the way imagine this is that users will be able to pick and choose just what they want. So maybe separate, specific accessories will be in a separate repos. Others that are similar, like all the fake accessories, can get a single repo, e.g. called `fakes`. I definitely agree that updating all of these will be worse, but I hope that the current rate of breaking changes will drop sooner or later.

cdce8p:
Wouldn't `setup/extras_require` be a way to achieve just that?

I hope as well that their aren't that many breaking changes any more, however until at least the async conversion is complete their might be more to come. Please don't get me wrong here, but I wouldn't want to create ten or more PRs just for one single breaking change.

ikalchev:
I totally understand - each repo will be maintained by its owner. Breaking changes were expected for me becaus there weren’t so many contributors until now. Once different points of view come, changes are needed to strike a balance and get the best of all. So everything is fine now I think.

As for the extras - this won’t work because it will not reduce package size and will require each third party accessory to add itself here, whereas in the showed example they can be developed independently

cdce8p:
> As for the extras - this won’t work because it will not reduce package size and will require each third party accessory to add itself here, whereas in the showed example they can be developed independently

You're right. Didn't though about that.

> Package size

The reason I suggested the change was to not include any additional accessories with `HAP-python`, since Home Assistant (and maybe other projects) don't need them. From a maintenance point of view, I would move the current `accessories` (those which should be kept up to date) to just one new repo. Those files aren't that big to begin with.

ikalchev:
So what do you hint, should I proceed with this merge (I need to add to the changelog first though).

cdce8p:
My comments where more about how we move forward after this.
Regarding the merge, I don't really have an opinion on it. We should just try to move to the final solution before the next release to limit breaking changes.

PR #117, opened 2018-05-07
Small changes - Accessory

cdce8p:
### Changes
- Removed unused method `accessory.create`.
- Removed `iid_manager` and `setup_id` parameter from `accessory` and `bridge` `init` calls.
- Improved `accessory.to_HAP()`

In theory this is a breaking change. However it's not very likely many people have used those parameters to begin with since the defaults are quite fine. Therefor this shouldn't be an issue.

Some changes are extracted from #104. I didn't wanted to just propose a single new line ;)
In addition this will help keep the other PR clean.

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -31,13 +31,7 @@ class Accessory:
 
     category = CATEGORY_OTHER
 
-    @classmethod
-    def create(cls, display_name, pincode, aid=STANDALONE_AID):
-        mac = util.generate_mac()
-        return cls(display_name, aid=aid, mac=mac, pincode=pincode)
-
-    def __init__(self, display_name, aid=None, mac=None, pincode=None,
-                 iid_manager=None, setup_id=None):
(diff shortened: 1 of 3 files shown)


pyhap/accessory.py
```diff
@@ -487,12 +479,7 @@ def to_HAP(self):
 
         .. seealso:: Accessory.to_HAP
         """
-        hap_rep = [super().to_HAP()]
```
ikalchev: Not sure if this is more readable or more efficient than the previous version.

cdce8p: I'm not 100% percent sure either. Just though about that line while I was working on the other PR and it looked pretty cool, especially since we already use list comprehensions in other places (mainly for the `to_HAP` methods actually).

At least it isn't worse than the before IMO, and it cuts five lines.

ikalchev: It's not worse, performance-wise I think. Just readability, but it's good enough, Merging

PR #118, opened 2018-05-08
Adding Code Style Checker

cdce8p:
Adding static code style checkers can hugely improve the code quality and catch errors sooner. At the moment these test fail completely, so we might have to decide if we want to add it to travis or leave it out for now. If we move forward with this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
-
 - Added `getter_callback` to Characteristics. [#90](https://github.com/ikalchev/HAP-python/pull/90)
 
 ### Changed
@@ -24,6 +23,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Developers
 - The `driver` event loop name changed from `event_loop` to `loop`. [#107](https://github.com/ikalchev/HAP-python/pull/107)
+- Added static code checks. To run them locally use `tox -e lint` and `tox -e pylint`. [#118](https://github.com/ikalchev/HAP-python/pull/118)
 
 
 


cdce8p:
I excluded some tests for now (mainly `docstrings`) and fixed many other lint errors. The ones that are still remaining require a bit more attention if we move forward.

https://travis-ci.org/ikalchev/HAP-python/jobs/376972170#L522-L569

ikalchev:
Sounds very good. The checks are currently passing because of this, can we exclude the failing checks and enable them gradually?

cdce8p:
What do you think about disabling `pylint` for `travis`, at least for now?
We can still test it locally with `tox`.

.travis.yml
```diff
@@ -8,11 +8,15 @@ branches:
     - /^\d\.\d+\.\d+(rc\d+|\.dev\d+)?$/
 matrix:
   include:
-    - python: 3.5
+    - python: "3.5"
+      env: TOXENV=lint
+    # - python: "3.5"
+    #   env: TOXENV=pylint
```
cdce8p: I disabled the `pylint` tests in `travis` for now, since they require more attention and should be handled separately.
The test suite can be run with `tox -e pylint`. That will exclude `missing-docs`.

tox.ini
```diff
@@ -1,28 +1,56 @@
...
   /usr/bin/make
   make
+
+
+[testenv:lint]
+basepython = {env:PYTHON3_PATH:python3}
+deps =
+  -r{toxinidir}/requirements_test.txt
+commands =
+  flake8 pyhap tests --ignore=D10,D205,D4,E501 --exclude=pyhap/accessories/*
```
cdce8p: The ignores should be removed once the docs pass.
Also the `--exclude` flag can be removed with #115.

tox.ini
```diff
@@ -1,28 +1,56 @@
...
+
+[testenv:pylint]
+basepython = {env:PYTHON3_PATH:python3}
+ignore_errors = True
+deps =
+  -r{toxinidir}/requirements_all.txt
+  -r{toxinidir}/requirements_test.txt
+commands =
+  pylint pyhap --disable=missing-docstring,invalid-name,fixme
```
cdce8p: Same here for the `--disabled` flag.

tox.ini
```diff
@@ -1,28 +1,56 @@
...
+
+[testenv:doc-errors]
+basepython = {env:PYTHON3_PATH:python3}
+ignore_errors = True
+deps =
+  -r{toxinidir}/requirements_all.txt
+  -r{toxinidir}/requirements_test.txt
+commands =
+  flake8 pyhap tests --select=D10,D205,D4,E501 --exclude=pyhap/accessories/*
+  pylint pyhap --disable=all --enable=missing-docstring
```
cdce8p: Run `tox -e doc-errors` locally to check which docs needs to be fixed.

cdce8p:
@ikalchev This should be ready now if you approve.

PR #124, opened 2018-05-21
WIP: Improvements 8 - Async helper methods

cdce8p:
## Changes:
* Changed order of execution, loop starts first (always)
* Execution will automatically stop on KeyBoardInterrupt
* Renamed `start` to `start_pyhap`
* Renamed `stop` to `stop_pyhap`
* Added asnyc helper methods
* Use helper functions for `run` and `stop` methods

---

@ikalchev The PR isn't finished yet, but it should run without issue. What do you think?


cdce8p:
This PR is ready for review now.

In addition to the helper methods, I combined all methods from `AsyncAccessory` with the `Accessory` so that the `async` version is no longer necessary. You can use which every variant you want, normal or async with `run` and `stop`, and `run_at_interval` will work with both versions as well.

If #125 is merged first, I would like to add an additional note with credit for the async helper methods.

pyhap/accessory_driver.py
```diff
@@ -493,23 +579,10 @@ def start(self):
...
-
-        logger.info("Starting event loop")
-        try:
-            self.loop.run_forever()
-        except KeyboardInterrupt:
-            pass
-        finally:
-            self.loop.close()
-            logger.info("Closed event loop.")
+        self.add_job(self.accessory.run)
```
ikalchev: Can you please add a note to the changelog that old implementations of Accessory.run, which still use the `while True: sleep (x); do_stuff()` will exhaust the thread pool with never-ending tasks

pyhap/accessory_driver.py
```diff
@@ -157,6 +181,68 @@ def __init__(self, *, address=None, port=51234,
...
+        Pyhap will be stopped gracefully on a KeyBoardInterrupt.
+        """
+        try:
+            logger.info('Starting the event loop')
+            self.add_job(self.start_pyhap)
+            self.loop.run_forever()
+        except KeyboardInterrupt:
+            self.loop.call_soon_threadsafe(
+                self.loop.create_task, self.async_stop())
+            self.loop.run_forever()
```
ikalchev: Why do we need a second `run_forever` (first one on line 192)?

pyhap/accessory_driver.py
```diff
@@ -157,6 +181,68 @@ def __init__(self, *, address=None, port=51234,
...
+            self.loop.call_soon(target, *args)
+        elif asyncio.iscoroutinefunction(target):
+            task = self.loop.create_task(target(*args))
+        else:
+            task = self.loop.run_in_executor(None, target, *args)
+
+        return task
+
+    @callback
+    def async_run_job(self, target, *args):
```
ikalchev: Can you elaborate on this? The idea here is to be able to execute a `@callback` immediately right? Why not let the client just run its immediate method if she knows it is not async.

pyhap/accessory_driver.py
```diff
@@ -450,7 +536,7 @@ def set_characteristics(self, chars_query, client_addr):
             chars_response.append(response)
         return {HAP_REPR_CHARS: chars_response}
 
-    def start(self):
+    def start_pyhap(self):
```
ikalchev: This is a "private" method right? Can we name it `_start` or `_do_start` (similar for `stop_pyhap`)?

accessories/TemperatureSensor.py
```diff
@@ -18,6 +16,6 @@ def __init__(self, *args, **kwargs):
         serv_temp = self.add_preload_service('TemperatureSensor')
         self.char_temp = serv_temp.configure_char('CurrentTemperature')
 
-    @AsyncAccessory.run_at_interval(3)
-    async def run(self):
+    @Accessory.run_at_interval(3)
+    def run(self):
```
ikalchev: Why not keep it `await def run`?

pyhap/accessory_driver.py
cdce8p: As long as it's not an async method that shouldn't be an issue, since it's basically the way it was before (just one never ending thread). However I agree that `run_at_interval` should be the preferred way. I'll add this.

pyhap/accessory_driver.py
cdce8p: In theory yes, however what if we want to change a method from `callback` to `async` or `normal`? We would need to change all calls as well. Calling these methods through the helpers, even for direct callback, simplifies that.

accessories/TemperatureSensor.py
cdce8p: `char.set_value` is still be blocking, so this should not be executed in the event loop.

pyhap/accessory_driver.py
cdce8p: The first one is so that `pyhap` doesn't stop, if all `run` methods are done. You should still be able to call `set_value` from a sensor readout for example or change a setting from the Home App.

The second one to run the `async_stop` method and I don't have to pass
```py
self.loop.call_soon_threadsafe(self.loop.create_task, self.async_stop())
```
to `self.loop.run_until_complete`. In the end both have the same effect.

pyhap/accessory_driver.py
cdce8p: I renamed and moved them upwards to the counterparts. In the long run, I plan to incorporated them into `async_start` (has to be created) and `async_stop` respectively.

pyhap/accessory_driver.py
```diff
@@ -157,6 +181,68 @@ def __init__(self, *, address=None, port=51234,
...
+    def start(self):
+        """Start the event loop and call `start_pyhap`.
+
+        Pyhap will be stopped gracefully on a KeyBoardInterrupt.
+        """
+        try:
+            logger.info('Starting the event loop')
+            self.add_job(self.start_pyhap)
+            self.loop.run_forever()
+        except KeyboardInterrupt:
```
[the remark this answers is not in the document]
cdce8p: > To be honest I haven't tested pyhap on windows. Ever. Have you?

No :smile: I only know that `KeyboardInterrupt` works for Windows as well.

> Signal handlers

`driver.stop` does still work, not just the `KeyboardInterrupt`. I don't have that much experience with running programs as `demon processes`, but I think HA doesn't use any signal handler and it still exits gracefully. Can you try stopping the `service` without any signal handlers?

> `driver.loop.add_signal_handler`

I would prefer the current solution is basically the same if not easier, since I don't need an extra method.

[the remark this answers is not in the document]
cdce8p: If the service exits gracefully even without the signal handler, I think we can remove `driver.signal_handler` as well.

pyhap/accessory_driver.py
ikalchev: Okay makes sense

pyhap/accessory_driver.py
ikalchev: Yes, the first call is clear, my issue was with the second one. Any chance something else submits to the loop so that the loop doesn't stop even after the async_stop finishes? In that case, isn't `run_until_complete` a better choice just to be on the safe side?

cdce8p: `async_stop` calls `loop.stop()` so it will terminate. Regarding other tasks, since the `Accessory.stop` methods can be `async` as well, it is even intended that other might add other calls to the event loop. The part missing here, which I will add later is to wait for all calls to finish before calling `stop`.

cdce8p:
No, all done. I just resolved the merge conflicts and added one small section to the file copyrights: https://github.com/ikalchev/HAP-python/pull/124/files#diff-9879d6db96fd29134fc802214163b95aR16

PR #125, opened 2018-05-23
Changes to organization files

cdce8p:
Some suggestion on for the overall structure of the `HAP-python` project. The original intend was to add a section to the `LICENSE` file giving credit to other open source projects when code was adopted.
Next I wanted to include the `LICENSE` file in the distribution and I ended out moving almost all static content to `setup.cfg`.

In addition to small style changes (4-space indent for `tox.ini` for better readability), I also added two scripts to simplify setup and release. The later now generates a build distribution as well, not just a source distro.

For `setup`, I modified some files. The `description-file` wasn't included at all, so I removed it (also not really necessary IMO). If you think otherwise, the `long_description` attribute should work. Furthermore I added `keywords` and extended the `classifier`.

## Changes
* Extended use of setup.cfg for static settings
* Moved version to pyhap.const
* Extended license information
* Added authors file
* Added scripts for setup and release

Tested with `setuptools 39.2.0`

## Disclaimer

I'm no legal export, so I can't guaranty that the additions to the `LICENSE` are the best way to go. My inspiration for this change was the `sandstorm` project. In their wiki they have a section [Giving credit to other open source projects](https://github.com/sandstorm-io/sandstorm/wiki/Giving-credit-to-other-open-source-projects) and I looked at their [LICENSE file](https://github.com/sandstorm-io/sandstorm/blob/master/LICENSE).

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+# Stop on errors
+set -e
+
+cd "$(dirname "$0")/.."
+
+# Install missing dependencies
+if [ -z "$(python3 -m pip list | grep 'wheel')" ]; then
+	python3 -m pip install wheel
+fi
+if [ -z "$(python3 -m pip list | grep 'twine')" ]; then
+	python3 -m pip install twine
+fi
+
+echo "====================================="
+echo "=   Generation source distribution  ="
+echo "====================================="
+python3 setup.py sdist
+
+echo "===================================="
+echo "=   Generation build distribution  ="
+echo "===================================="
+python3 setup.py bdist_wheel
+
+echo "====================="
+echo "=   Upload to pypi  ="
+echo "====================="
+python3 -m twine upload dist/* --skip-existing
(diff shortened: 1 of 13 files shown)


scripts/setup
```diff
@@ -0,0 +1,17 @@
...
+
+echo "=============================="
+echo "=   Installing dependencies  ="
+echo "=============================="
+pip install -e .[qrcode]
+
+echo "==================================="
+echo "=   Installing test dependencies  ="
+echo "==================================="
+python3 -m pip install tox
```
ikalchev: Can we add this in the test requirements?

scripts/release
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
```
ikalchev: Can you please add a one-liner comment explaining the purpose of this script and the setup script.

cdce8p:
> Can we add this in the test requirements?

:+1: 

> Can you please add a one-liner comment explaining the purpose of this script and the setup script.

No problem, must have slipped through the first time.

PR #130, opened 2018-05-29
Fix spelling error, bump version to 2.2.2

cdce8p:
Unfortunately I had wrote `date` instead of `data`, which resulted in the wheel being broken again :disappointed:

The upside is that it should work now, which I could test, since I just found out that I had to delete the build directory before I start a new one.

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -6,6 +6,14 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+# Remove previous build
+if [ -n "$(ls | grep 'build')" ]; then
+	rm -r build/
+fi
+
+# Upgrade setuptools to the latest version
+python3 -m pip install --upgrade setuptools
+
 # Install missing dependencies
 if [ -z "$(python3 -m pip list | grep 'wheel')" ]; then
 	python3 -m pip install wheel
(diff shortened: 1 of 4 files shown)


scripts/release
```diff
@@ -6,6 +6,14 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+# Remove previous build
+if [ -n "$(ls | grep 'build')" ]; then
+	rm -r build/
+fi
+
+# Upgrade setuptools to the latest version
+python3 -m pip install --upgrade setuptools
```
ikalchev: What is the reason to add this?

cdce8p: I noticed the the metadata `project_urls` weren't redenered correctly and that the tag `bdist_wheel (0.30.0)` was included. That led me to belive that the `setuptools` version is probably outdated. Since this happend to me as well, I decided to add this.

If you want to, I'll remove it.
Just running `pip install --upgrade setuptools` would be enough as well.

ikalchev: This may unexpectedly change the user env, I would prefer not to do it, if it is OK for you

cdce8p: Absolutely! In that case it might be best to remove the `wheel` and `twine` installs down below as well.

cdce8p:
I've removed all `pip` commands from the `release` script. Would be great if you could update `setuptools` nevertheless (only if it isn't the current version `39.2.0` already).

ikalchev:
Unfortunately I will not be able to do another release for a few hours

cdce8p:
No problem. It was my fault to begin with.

PR #142, opened 2018-07-27
add support for "lightbulbs" connected to Hue

lobeck:
This is a basic accessory discovering lightbulbs connected to a Hue Bridge and exposes them to HomeKit.

### DIFF
--- a/accessories/Hue.py
+++ b/accessories/Hue.py
@@ -1,0 +1,88 @@
+"""Basic Hue integration
+Allows control of incompatible Accessories like Osram Lightify Plugs (Osram Smart+)
+
+I have no idea if this is a good implementation, but it works for me :D
+"""
+import logging
+import signal
+import certifi
+import requests
+import urllib3
+
+from pyhap.accessory import Bridge, Accessory
+from pyhap.accessory_driver import AccessoryDriver
+from pyhap.const import CATEGORY_LIGHTBULB
+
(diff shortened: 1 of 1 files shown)


accessories/Hue.py
```diff
@@ -0,0 +1,88 @@
...
+import signal
+import certifi
+import requests
+import urllib3
+
+from pyhap.accessory import Bridge, Accessory
+from pyhap.accessory_driver import AccessoryDriver
+from pyhap.const import CATEGORY_LIGHTBULB
+
+logging.basicConfig(level=logging.INFO)
```
ikalchev: Maybe leave this to the module that starts HAP, e.g. the main.py

accessories/Hue.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self, *args, id, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        serv_light = self.add_preload_service('Lightbulb')
+        self.char_on = serv_light.configure_char(
+            'On', setter_callback=self.set_bulb)
+
+        self.id = id
+
+    def __setstate__(self, state):
```
ikalchev: This method is not needed now (if you don't plan to pickle the accessory)

accessories/Hue.py
ikalchev: If you decide to remove it, can you use `logger = logging.getLogger(__name__)`

PR #144, opened 2018-08-18
NeoPixelsLightStrip

LasCabos:
Added ability to interface with NeoPixels RGB LEDs. This will allow changing of Color, Brightness and State(On/Off) of the NeoPixels attached to the RPi.
This accessory uses all available HomeKit Lightbulb API services (Hue, Saturation, Brightness).
This commit requires the installation of rpi_ws281x library. The procedure is outlined in # https://learn.adafruit.com/neopixels-on-raspberry-pi/software.

### DIFF
--- a/accessories/NeoPixelLightStrip.py
+++ b/accessories/NeoPixelLightStrip.py
@@ -1,0 +1,148 @@
+# An Accessory for Adafruit NeoPixels attached to GPIO Pin18
+# Tested using Python 3.6 Raspberry Pi
+# This device uses all available services for the Homekit Lightbulb API
+# Note: set your neopixels settings under the #NeoPixel constructor arguments
+# Note: RPi GPIO must be PWM. Neopixels.py will warn if wrong GPIO is used
+#       at runtime
+# Note: This Class requires the installation of rpi_ws281x lib
+#       Follow the instllation instructions;
+#           git clone https://github.com/jgarff/rpi_ws281x.git
+#           cd rpi_ws281x
+#           scons
+#
+#           cd python
+#           sudo python3.6 setup.py install
+# https://learn.adafruit.com/neopixels-on-raspberry-pi/software
+
+# Apple Homekit API Call Order
+# User changes light settings on iOS device
+# Changing Brightness - State - Hue - Brightness
+# Changing Color      - Saturation - Hue
+# Changing Temp/Sat   - Saturation - Hue
+# Changing State      - State
+
+# import logging
+from neopixel import *
+
+from pyhap.accessory import Accessory
+from pyhap.const import CATEGORY_LIGHTBULB
+
+
+class NeoPixelLightStrip(Accessory):
+
+    category = CATEGORY_LIGHTBULB
+
+    __accessoryState = 0  # State of the neo light On/Off
+    __hue = 0  # Hue Value 0 - 360 Homekit API
+    __saturation = 100  # Saturation Values 0 - 100 Homekit API
+    __brightness = 100  # Brightness value 0 - 100 Homekit API
+
+    # NeoPixel constructor arguments
+    LED_COUNT = 8
+    LED_PIN = 18
+    LED_FREQ_HZ = 800000
+    LED_DMA = 10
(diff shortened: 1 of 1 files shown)


accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+#           cd python
+#           sudo python3.6 setup.py install
+# https://learn.adafruit.com/neopixels-on-raspberry-pi/software
+
+# Apple Homekit API Call Order
+# User changes light settings on iOS device
+# Changing Brightness - State - Hue - Brightness
+# Changing Color      - Saturation - Hue
+# Changing Temp/Sat   - Saturation - Hue
+# Changing State      - State
```
cdce8p: Instead of using single line comments, maybe use python doc comments: `"""" """`

accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+
+# import logging
```
cdce8p: That should be removed if not needed.

accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+            'On', setter_callback=self.set_bulb)
+        self.char_on = serv_light.configure_char(
+            'Brightness', setter_callback=self.set_brightness)
+
+        # Must be called before any colors can be applied to neoPixels
+        self.__neo_strip.begin()
+
+    # def __setstate__(self, state):
+    #     print("___ setstate ___")
+    #     self.__dict__.update(state)
```
cdce8p: Same for this method

accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def set_brightness(self, value):
+        self.__brightness = value
+        self.set_hue(self.__hue)
+
+    def set_saturation(self, value):
+        self.__saturation = value
+        self.set_hue(self.__hue)
+
+    def Update_NeoPixel_With_Color(self, red, green, blue):
```
cdce8p: Python methods should be lowercase: `update_neopixel_with_color`

accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+        # or it could be the neopixel.py library
+        # Change the setPixelColor inputs for yourself below
+        for i in range(self.LED_COUNT):
+            self.__neo_strip.setPixelColor(
+                i, Color(int(green), int(red), int(blue)))
+
+        self.__neo_strip.show()
+
+    def stop(self):
+        super().stop()
```
cdce8p: This shouldn't be necessary

accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def hsv_to_rgb(self, h, s, v):
+        # This function takes
+        # h - 0 - 360 Deg
+        # s - 0 - 100 %
+        # v - 0 - 100 %
```
cdce8p: Also a python doc comment

accessories/NeoPixelLightStrip.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class NeoPixelLightStrip(Accessory):
+
+    category = CATEGORY_LIGHTBULB
+
+    __accessoryState = 0  # State of the neo light On/Off
+    __hue = 0  # Hue Value 0 - 360 Homekit API
+    __saturation = 100  # Saturation Values 0 - 100 Homekit API
+    __brightness = 100  # Brightness value 0 - 100 Homekit API
```
cdce8p: Those should probably be instance vars. Also local vars in python only have one leading underscore.

cdce8p:
Left a few comments that should improve your implementation

ikalchev:
I agree with the comments by @cdce8p, however, as this is not submitted as part of the framework itself - let's merge it.

@LasCabos you can have a look at the README in pyhap/accessories if you want to distribute this accessory more easily. Thanks!

PR #162, opened 2018-10-16
Char - ignore duplicate values

cdce8p:
`HomeKit` automations initiate calls to the server without taking the current value into account. That isn't the case normally and even scenes respect them. I think the best option would be to integrate a check in `client_update_value` if the value is already set and if so ignore that call.

See also: https://github.com/home-assistant/home-assistant/pull/17453

CC: @adrum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Sections
 ### Developers
 -->
 
-## [2.2.3] - 2018-10-05
+## [2.x.x] - TBA
 
 ### Added
 - Added `NeoPixelsLightStrip` accessory. [#144](https://github.com/ikalchev/HAP-python/pull/144)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,14 +16,15 @@ Sections
...
-## [2.2.3] - 2018-10-05
+## [2.x.x] - TBA
 
 ### Added
 - Added `NeoPixelsLightStrip` accessory. [#144](https://github.com/ikalchev/HAP-python/pull/144)
 - Added new Accessory categories. [Commit](https://github.com/ikalchev/HAP-python/commit/dbf1d5d8fea814af52098f3a2f3cd5a47cd889c1)
 
 ### Changed
 - Spelling fix - executor (accessory_driver). [#159](https://github.com/ikalchev/HAP-python/pull/159)
+- Char.update_client_value now ignores call if value is already set. This could happen during automations. [#162](https://github.com/ikalchev/HAP-python/pull/162)
```
ikalchev: Char.client_update_value

cdce8p:
No, target and current state are both set to close, so this shouldn't be the issue.

I believe that it's an inconsistency (planned/unplanned) with the Home App. Normally you can't close something that's already closed in the Home App, but this check seems to be missing for `scenes` and `automations`.

PR #165, opened 2018-10-24
Add a fake WindowCovering accessory script

lboue:
Should I remove the 2 following function too?

- def set_position_state(self, value)
- def set_current_position(self, value)

### DIFF
--- a/shutter.py
+++ b/shutter.py
@@ -1,0 +1,202 @@
+"""Starts a fake fan, lightbulb, garage door and a TemperatureSensor
+"""
+import logging
+import signal
+import random
+
+from pyhap.accessory import Accessory, Bridge
+from pyhap.accessory_driver import AccessoryDriver
+from pyhap.const import (CATEGORY_FAN,
+                         CATEGORY_LIGHTBULB,
+                         CATEGORY_GARAGE_DOOR_OPENER,
+                         CATEGORY_WINDOW_COVERING,
+                         CATEGORY_SENSOR)
+
+'''
+# ### CATEGORY values ###
+# Category is a hint to iOS clients about what "type" of Accessory this
+# represents, for UI only.
+CATEGORY_OTHER = 1
+CATEGORY_BRIDGE = 2
+CATEGORY_FAN = 3
+CATEGORY_GARAGE_DOOR_OPENER = 4
+CATEGORY_LIGHTBULB = 5
+CATEGORY_DOOR_LOCK = 6
+CATEGORY_OUTLET = 7
+CATEGORY_SWITCH = 8
+CATEGORY_THERMOSTAT = 9
+CATEGORY_SENSOR = 10
+CATEGORY_ALARM_SYSTEM = 11
+CATEGORY_DOOR = 12
+CATEGORY_WINDOW = 13
+CATEGORY_WINDOW_COVERING = 14
+CATEGORY_PROGRAMMABLE_SWITCH = 15
+CATEGORY_RANGE_EXTENDER = 16
+CATEGORY_CAMERA = 17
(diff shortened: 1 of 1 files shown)


shutter.py
```diff
@@ -0,0 +1,196 @@
...
+        
+        # Add the WindowCovering service. Also add optional characteristics to it.
+        serv_cover = self.add_preload_service(
+            'WindowCovering', chars=['ObstructionDetected'])
+
+        self.char_target_pos = serv_cover.configure_char(
+            'TargetPosition', setter_callback=self.set_target_position)
+            
+        self.char_state = serv_cover.configure_char(
+            'PositionState', setter_callback=self.set_position_state)
```
cdce8p: You don't need to assign this char a local variable. The `setter_callback` is never called, so you can remove it. Maybe add `value=2` here as a parameter.

shutter.py
```diff
@@ -0,0 +1,196 @@
...
+            
+        self.char_cur_pos = serv_cover.configure_char(
+            'CurrentPosition', setter_callback=self.set_current_position)
```
cdce8p: The `setter_callback` is also never called, so it can be removed.

shutter.py
```diff
@@ -0,0 +1,196 @@
...
+                                     
+        #serv_cover.configure_char('PositionState', value=2) # STOPPED
+        #serv_cover.configure_char('ObstructionDetected', value=1)
+                                     
+    def set_target_position(self, value):
+        logging.info("WindowCovering TargetPosition value: %s", value)
+        self.get_service('WindowCovering')\
+            .get_characteristic('TargetPosition')\
+            .set_value(value)
```
cdce8p: Don't set the char to the value that it posted. This isn't necessary. Just remove it.

shutter.py
```diff
@@ -0,0 +1,196 @@
...
+
+        time.sleep(5)   # Delays for 5 seconds closing the shutter.
+
+        logging.info("WindowCovering CurrentPosition value: %s", value)
+        self.get_service('WindowCovering')\
+            .get_characteristic('CurrentPosition')\
+            .set_value(value)
```
cdce8p: You've assigned `CurrentPosition` a local var. Why not use it:
```py
self.char_cur_pos.set_value(value)
```

shutter.py
```diff
@@ -0,0 +1,196 @@
...
+        logging.info("WindowCovering PositionState value: %s", value)
+        self.get_service('PositionState')\
+            .get_characteristic('PositionState')\
+            .set_value(value)
+
+    def set_current_position(self, value):
+        logging.info("WindowCovering CurrentPosition value: %s", value)
+        self.get_service('WindowCovering')\
+            .get_characteristic('CurrentPosition')\
+            .set_value(value)
```
cdce8p: The methods `set_position_state` and `set_current_position` can be removed, since they are never called.

shutter.py
```diff
@@ -0,0 +1,196 @@
...
+    category = CATEGORY_WINDOW_COVERING
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        """Quick assign basic accessory information."""
+        self.set_info_service(firmware_revision=2,  manufacturer="Brand",
+                model="Shutter", serial_number="0123456789")
+        
+        # Add the WindowCovering service. Also add optional characteristics to it.
```
cdce8p: This comment and the one about assigning accessory infos can be removed as well.

cdce8p:
@ikalchev I don't like the idea of merging a PR that isn't finished yet, only to finish it later. Instead consider committing the changes directly to the PR branch. Maintainers do generally have the permission to do so.

lboue:
OK. I will fix all that items ASAP.

PR #168, opened 2018-11-02
Temp disable 'update_advertisment' after pairing

cdce8p:
In some cases the pairing doesn't work, because unregister and register of the mDNS service doesn't work as expected. If we remove the call to `update_advertisement` inside the `pair` and `unpair` methods, we eliminate that point of failure. The only drawback is that the user would either need to restart HAP-python after unpairing (that's want I would propose) or the service would need to be discoverable constantly.

Related issues:
https://github.com/home-assistant/home-assistant/issues/17181
https://github.com/home-assistant/home-assistant/issues/14567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,13 +21,13 @@ Sections
 ### Added
 - Added a `asyncio.SafeChildWatcher` as part of `AccessoryDriver.start`.
 - Added `Camera.stop`, which terminates all streaming processes.
+- `AccessoryDriver.safe_mode` parameter. Set with `driver.safe_mode = True` before `driver.start` to disable `update_advertisement` call for `pair` and `unpair`. After unpairing a restart is necessary. [#168](https://github.com/ikalchev/HAP-python/pull/168)
 
 ### Changed
 - The default implementations of the `Camera`'s `start_stream`, `stop_stream` and
 `reconfigure_stream` are now async.
 - The streaming process is started with `asyncio.create_subprocess_exec` instead of
 `subprocess.Popen`
-- Temporary disabled `update_advertisement` call after `pair` and `unpair`. To pair the accessory again after a successful unpair, you need to restart `HAP-python`. [#167](https://github.com/ikalchev/HAP-python/pull/167)
 
 ### Fixed
 - `AcessoryDriver.add_job` now correctly schedules coroutines wrapped in functools.partial.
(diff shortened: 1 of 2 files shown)


ikalchev:
Hi @cdce8p,

How about we go for a more optimistic approach (Why I think we already discussed this somewhere? Has this happened before?):

```python
try:
   self.update_advertisement()
except NonUniqueNameException:
   logging.error(whatever)
```

I think that I put this: https://github.com/ikalchev/HAP-python/blob/dev/pyhap/accessory_driver.py#L449
because of the same reason. So in the above code it may also be worth it to defer the update to a bit or sleep longer between unreg and reg.

The reason I want to go in this direction is because this issue is **not** occurring for most people, so we shouldn't change the behavior for them.

What do you think?

Best,
Ivan

cdce8p:
Yeah, I proposed something similar with #72 with `try / except`.
However I believe now that this wouldn't work, which is why I removed the methods completely. The reason being that the error is happening **after** the service is unregistered and during the new registration. So when HomeKit tries to confirm the pairing, it fails.

I though about what you said about not effecting everybody else.
What do you think about introducing a `safe_mode` parameter that would disable the calls to `update_advertisement` if set? That way we wouldn't change anything during normal use, I can easily implement this in Home Assistant and don't have to suggest replacing the `accessory_driver` file every time.

My implementation would look something like this:
```py
class AccessoryDriver:
  def __init__(self, *, ...):
    self.safe_mode = False

  def pair(self, ...):
    ...
    if not self.safe_mode:
      self.update_advertisement()

  def unpair(self, ...):
    ...
    if not self.safe_mode:
      self.update_advertisement()
```

Instantiating the `AccessoryDriver` wouldn't change. Activating `safe_mode` would happen like this:
```py
driver = AccessoryDriver(...)
driver.safe_mode = True
...
driver.start()
```

ikalchev:
Well, I guess it would be best for HA. I would propose something like

```python
async def update_advertisement(self):
    self.advertiser.unregister_service(self.mdns_service_info)
    self.mdns_service_info = AccessoryMDNSServiceInfo(
            self.accessory, self.state)
    await time.sleep(0.1)  # Doing it right away can cause crashes.
    try:
        self.advertiser.register_service(self.mdns_service_info)
    except NonUniqueNameException:
        self.loop.call_later(2, self.update_advertisement)
```

But then to make it work properly, we would need to implement state transitions so that another `update_advertisment` is not triggered, etc, etc... I can give it a try at some point.

My only request would be to put a reference to one of the issues, so that we document what required this parameter.

cdce8p:
I'll add the issue reference.

Regarding your suggestion: It would cause the pairing to fail, since the Home app expects the mDNS service. In your example it would be unavailable for at least two seconds.

PR #177, opened 2018-12-07
Also consider EBADF to be a timeout. Add some logging.

xbtsw:
I have been using HAP-python with home assistant on a Raspberry Zero W. My homekit will randomly goes into "No Response" status and also randomly recover on itself. After some debugging it looks like the socket sometimes will throw exception with `errno.EBADF`, at this point HAP-python will also throw and my homekit will immediately goes into "No Response" until next `pair-verify` initiated by homekit hub.

I don't exactly know in which case the `EBADF` exception will happen, but my guess is that Raspberry ZW only have 2.4GHz wifi, and my area's band is very crowded, so maybe when there's interference the `EBADF` will be the error number.

After applied this change locally, I has been running without "No Response" for about two weeks.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -780,7 +780,7 @@ class HAPServer(socketserver.ThreadingMixIn,
                      b"Content-Length: "
 
     TIMEOUT_ERRNO_CODES = (errno.ECONNRESET, errno.EPIPE, errno.EHOSTUNREACH,
-                           errno.ETIMEDOUT, errno.EHOSTDOWN)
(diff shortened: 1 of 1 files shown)


ikalchev:
Hi Wei, 

Thanks for taking time debugging this! I think this will solve #176. What worries me is that EBADF should happened because of e.g. a read after the socket is closed. This means that there is either a race or the socket state is not managed properly

Anyway, I think we can take this change until we narrow down why EBADF happens in the first place.

Thanks for this!

Best,
Ivan

PR #198, opened 2019-06-20
Update hap_server.py

a3135134:


bdraco:
@a3135134 This caused breakage for me, but I was able to solve it.  For additional details see https://github.com/ikalchev/HAP-python/pull/214

sander-vd:
Can confirm this works, so also solves my issue when reconnecting wifi #221 thanks bdraco!

bdraco:
Should be solved in 2.7.0+

ikalchev:
Done in #216

Issue #197, opened 2019-06-20
HTTP server must be HTTP/1.1


a3135134:
According to HAP specification 5.2.4, HTTP server must be HTTP/1.1. Now it's 1.0.

If the controller doesn't maintain a persistent connection, the pair verify progress will raise an error. And HTTP 1.0 doesn't require a persistent connection by default.

I just added the code
self.protocol_version = "HTTP/1.1" 
in __init__() function of  class HAPServerHandler in "/pyhap/hap_server.py". And it works.

schinckel:
This seems to have fixed the issues I've noticed recently too, where my HAP-python bridge would drop-off fairly frequently.

schinckel:
...may have spoken too soon... :(

bdraco:
@schinckel  Does https://github.com/bdraco/HAP-python/commit/fbabe83c15f08c01654f5acac2e9e53a7aeeae05 solve this for you?

schinckel:
I’m not sure that the problems I was seeing were application-related: I think it was a configuration issue, where reflected advertisements caused the zeroconf library to stop advertising.

PR #214, opened 2019-12-02
Add support for HTTP/1.1 and random non-responsive devices #197

bdraco:
Homekit needs the headers and content sent
in a single write call or it will randomly
get connection reset by peer which leads
to iOS stalling and devices showing up
as non-responsive.

Additionally we avoid connection resets
by draining any incoming data from the buffer
when we close the connection.

Without HTTP/1.1 keep-alives, 50+ devices
would cause a bridge to timeout or devices
to go non-responsive because homekit would
have to make a new connection to get / put
for each device.  With HTTP/1.1 it can avoid
all of this overhead

After this change, devices in homekit on
the bridges with > 80 devices were
now immediately responsive in testing.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -24,6 +24,7 @@ import ed25519
 
 import pyhap.tlv as tlv
 from pyhap.util import long_to_bytes
+from pyhap.const import __version__
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


bdraco:
I think I can simplify this a bit more. Will resubmit later

bdraco:
Replaced with #216

PR #215, opened 2019-12-04
Ensure events are not sent to the initiator client

bdraco:
Clients that made the characteristic change are NOT supposed to get
events
about the characteristic change as it can cause an unexpected HTTP
disconnect (connection
reset by peer) which leads temporary unresponsive accessories and
violates the HAP spec

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -294,7 +294,7 @@ class Accessory:
 
     # Driver
 
-    def publish(self, value, sender):
+    def publish(self, value, sender, sender_client_addr=None):
         """Append AID and IID of the sender and forward it to the driver.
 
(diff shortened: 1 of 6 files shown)


ikalchev:
@bdraco You said that the fix in #197 likely fixes this one as well. Is this then still needed? Or are the changes here needed to make pyhap HAP-spec compliant?

bdraco:
@ikalchev  The changes are still needed to make it compliant and avoid the disconnect when a message is sent when the client is not expecting one.

bdraco:
The easiest way I found to replicate the disconnect is to rapidly turn a light on/off that is linked to home assistant as it will cause the event message to be sent to the server from the client while the client is writing.

PR #216, opened 2019-12-04
Add support for HTTP/1.1 and random non-responsive devices #197

bdraco:
Homekit needs the headers and content sent
in a single write call or it will randomly
get connection reset by peer which leads
to iOS stalling and devices showing up
as non-responsive.

Without HTTP/1.1 keep-alives, 50+ devices
would cause a bridge to timeout or devices
to go non-responsive because homekit would
have to make a new connection to get / put
for each device.  With HTTP/1.1 it can avoid
all of this overhead

After this change, devices in homekit on
the bridges with > 80 devices were
now immediately responsive in testing.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -24,6 +24,7 @@ import ed25519
 
 import pyhap.tlv as tlv
 from pyhap.util import long_to_bytes
+from pyhap.const import __version__
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pyhap/hap_server.py
```diff
@@ -190,9 +198,24 @@ def _upgrade_to_encrypted(self):
...
+        # The below code does what end_headers does internally
+        # except it combines the headers and the content
+        # into a single write instead of two calls to
+        # self.wfile.write
+        #
+        # TODO: Is there a better way that doesn't involve
+        # touching _headers_buffer ?
+        #
+        self.connection.sendall(b"".join(self._headers_buffer) + b"\r\n" + bytesdata)
+        self._headers_buffer = []
```
ikalchev: Is this really needed? I.e. how often do you see a stall if this is not done?

ikalchev:
One more thing - which python did you test with?
PS: Sorry for the delay on this, I hope we can merge and release this by end of the week.

pyhap/hap_server.py
bdraco: I would see it drop the connection and then end up in a stall state for a while until iOS hit an internal logic to reconnect.  It would happen about once a week and I'd usually have to reboot the phone to make everything work again.  This fix is similar to an issue I've seen before with apple's webdav library.

bdraco: The underlying issue is likely a bug in apple's iOS libraries that isn't dealing well with data coming at unexpected times (reading when its expecting to write and vise-versa).  It is also likely the reason #215 also causes disconnects.  Apple developer support says its a "known" issue that is being worked on.

ikalchev: Hey thanks for the insights, sounds good!

ikalchev:
Thank you for the change, I will release the new version in pip over the weekend.

Issue #190, opened 2019-04-27
Hundreds of threads after day or more of activity


undersun81:
Currently I have HAP server running for about 2 weeks and any request just fails because it can't create a new thread.

gdb attach <pid>
gdb info thread 
Shows 254 threads, mostly in recv for example
Thread 0x553c7470 (LWP 7985) "python3" 0x76f3e850 in recv () at ../sysdeps/unix/syscall-template.S:84

undersun81:
I think I have a good guess about what is going on here.
Several Apple devices I have at home (iPhones, iPad and HomePod) periodically or by a user request ask HAP-server about state or subscribe. From logs I can see they use different client ports, but they(clients) never close these sockets and next time use another port and HAP-server creates new connection and new corresponding thread. Linux by default has very long timeout values for TCP-sockets, so eventually I have hundreds of threads and open connections.  So I hit the limit somewhere in python and it just fails  with error when trying to create new thread. This story happen to me every evening, then during night most of sockets timed out and I had proper usage during the day, and story repeated in the evening.

HAP uses blocking sockets, so my solution was to set timeout for a client socket when we connected.
So in hap_server.py in method HAPServer.get_request I added 
"client_socket.settimeout(60)" after "client_socket, client_addr = super(HAPServer, self).get_request()".

This change has fixed my problem, but I'm not an expert in HomeKit protocol and not sure about subscription case. May be clients which subscribed expect to hold connection as long as possible. If so then they should send some ping periodically into the same connection and not just start talking through another socket. Looks like apple devices just don't care to properly shutdown those sockets (may be in some specific case, like sleeping device) and they seem don't care if we have short timeout on them.

niobos:
I don't think it has to do with open connections. I have TCP-keepalive configured, so dead connections get killed after about 5 minutes. Still, I see the number of threads increasing over time...

PR #222, opened 2019-12-11
Assign a timeout for client sockets of HAPServer.

vladimirlagunov:
Sometimes TCP connections are being lost. For instance, when you move your phone out of a Wi-Fi network, the phone is not able to close the TCP connection with the server and the connection remains open. After a while, an operating system starts refusing to create a new thread: `RuntimeError: can't start new thread`.

This commit introduces a timeout for client sockets. Now, if you move your phone out of a Wi-Fi, a connection between the phone and the server will be closed in a five minutes.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -782,6 +782,8 @@ class HAPServer(socketserver.ThreadingMixIn,
     TIMEOUT_ERRNO_CODES = (errno.ECONNRESET, errno.EPIPE, errno.EHOSTUNREACH,
                            errno.ETIMEDOUT, errno.EHOSTDOWN, errno.EBADF)
 
+    CLIENT_SOCKET_TIMEOUT = 300
+
     @classmethod
     def create_hap_event(cls, bytesdata):
         """Creates a HAP HTTP EVENT response for the given data.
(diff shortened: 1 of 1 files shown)


werehuman:
Fixes #190

bdraco:
@werehuman HomeKit is expecting to keep the connection open for days at a time without the server dropping the connection.

There were a few bugs in the connection handling where we didn't handle the connection being dropped by the client that have since been fixed in 2.8.4.  Can you retest and see if its fixed without having to add a timeout?

werehuman:
@bdraco I managed to update my HomeAssistant and I've been testing HAP-python==2.9.1 for more than a two weeks. Of course, as it always happens, upgrade of HomeAssistant brought a lot of suffering, I had to tune everything from scratch again.

Now it can't work even a day without maintenance. Before update my patched version HAP-python had been working without any problem for half a year. Right after I've updated, the described bug returned with exactly the same symptoms.

bdraco:
We fixed another protocol level issue in 2.9.2. Can you try that version?

bdraco:
There are also some thread fixes in Home Assistant 0.113 (not yet released) that may be a contributing factor.

bdraco:
I think this can be closed now

PR #229, opened 2020-03-19
Add support for service callbacks

bdraco:
For some services we want to send all the char value
changes at once.  This resolves an issue where we send
ON and then BRIGHTNESS and the light would go to 100%
and then dim to the brightness because each callback
would only send one char at a time. 

For additional details about why this is needed:
https://github.com/home-assistant/core/pull/32348
https://github.com/home-assistant/core/issues/13695
https://github.com/home-assistant/core/issues/13768
https://github.com/home-assistant/core/issues/32278

The end result is that home assistant can now get sets for a service in one call and do the right thing
```
2020-03-19 02:14:28 DEBUG (Thread-48) [custom_components.homekit.type_lights] _set_chars: {'Brightness': 45}
2020-03-19 02:14:29 DEBUG (Thread-48) [custom_components.homekit.type_lights] _set_chars: {'On': 1, 'Brightness': 25}
2020-03-19 02:14:30 DEBUG (Thread-48) [custom_components.homekit.type_lights] _set_chars: {'On': 1, 'Brightness': 66}
```

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -54,6 +54,8 @@ logger = logging.getLogger(__name__)
 
 CHAR_STAT_OK = 0
 SERVICE_COMMUNICATION_FAILURE = -70402
+SERVICE_CALLBACK = 0
+SERVICE_CALLBACK_DATA = 1
 
 
 def callback(func):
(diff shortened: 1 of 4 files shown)


bdraco:
@ikalchev  Hope you are staying safe.  Any chance you could look at this one this weekend?

ikalchev:
Overall it makes sense. We need to make sure to properly document the fact that you can register callbacks in both the service and the char, which can cause some confusion.

bdraco:
Good call. Will send a PR to update the docs.

Issue #230, opened 2020-03-19
AssertionErrors referencing macOS machine


jjlawren:
When using the Home Assistant [Homekit integration](https://www.home-assistant.io/integrations/homekit/), I occasionally get backtraces referencing HAP-python:
```
----------------------------------------
Exception happened during processing of request from ('192.168.2.2', 64896)
Traceback (most recent call last):
  File "/usr/lib/python3.7/socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "/srv/homeassistant-3.7/lib/python3.7/site-packages/pyhap/hap_server.py", line 880, in finish_request
    self, self.accessory_handler)
  File "/srv/homeassistant-3.7/lib/python3.7/site-packages/pyhap/hap_server.py", line 151, in __init__
    super(HAPServerHandler, self).__init__(sock, client_addr, server)
  File "/usr/lib/python3.7/socketserver.py", line 720, in __init__
    self.handle()
  File "/usr/lib/python3.7/http/server.py", line 428, in handle
    self.handle_one_request()
  File "/usr/lib/python3.7/http/server.py", line 394, in handle_one_request
    self.raw_requestline = self.rfile.readline(65537)
  File "/usr/lib/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/srv/homeassistant-3.7/lib/python3.7/site-packages/pyhap/hap_server.py", line 702, in recv_into
    data = self.recv(nbytes, flags)
  File "/srv/homeassistant-3.7/lib/python3.7/site-packages/pyhap/hap_server.py", line 725, in recv
    block_length_bytes = self.socket.recv(self.LENGTH_LENGTH)
  File "/srv/homeassistant-3.7/lib/python3.7/site-packages/pyhap/hap_server.py", line 713, in recv
    assert not flags and buflen > self.LENGTH_LENGTH
AssertionError
----------------------------------------
```
In the backtrace above, the 192.168.2.2 host is an unrelated macOS host running 10.15.3. That host has Homekit configured via the Home app but otherwise doesn't provide any Homekit services.

bdraco:
@jjlawren Is this still happening in 0.108 ?

jjlawren:
I haven't seen it since upgrading to 0.108.X.

jjlawren:
@bdraco Sorry I spoke too soon. It happened again on 0.108.3 this morning.

jjlawren:
[...]

PR #231, opened 2020-03-22
Home assistant camera

xdissent:
These are changes required for integrating camera support into Home Assistant core. See home-assistant/core#32527

PR #235, opened 2020-04-04
Avoid disconnect on short read

bdraco:
Handle short read in crypto stream
    
*  This implements the TODO to handle the short read.
    
*  recv_into should be a bit more efficient as well.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -733,19 +733,12 @@ class HAPSocket:
                     # It may be that we already read some data and we have
                     # 1 byte left, return whatever we have.
                     return result
-                block_length_bytes = self.socket.recv(self.LENGTH_LENGTH)
+                # Always wait for a full block to arrive
+                block_length_bytes = self.socket.recv(
+                    self.LENGTH_LENGTH, flags=socket.MSG_WAITALL
+                )
                 if not block_length_bytes:
                     return result
-                if len(block_length_bytes) != self.LENGTH_LENGTH:
-                    # Handle a short read where we only get
-                    # one bytes.  This should be very rare, but it happens.
-                    block_length_bytes_remainder = self.socket.recv(
-                        self.LENGTH_LENGTH - len(block_length_bytes)
-                    )
-                    if not block_length_bytes_remainder:
-                        return result
-                    block_length_bytes += block_length_bytes_remainder
-                    assert len(block_length_bytes) == self.LENGTH_LENGTH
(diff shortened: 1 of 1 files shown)


bdraco:
Not urgent as its rare.  Likely could go with #231

bdraco:
May fix #230

pyhap/hap_server.py
```diff
@@ -737,8 +736,16 @@ def recv(self, buflen=1042, flags=0):
...
-                # TODO: handle this
-                assert len(block_length_bytes) == self.LENGTH_LENGTH
+                if len(block_length_bytes) != self.LENGTH_LENGTH:
+                    # Handle a short read where we only get
+                    # one bytes.  This should be very rare, but it happens.
+                    block_length_bytes_remainder = self.socket.recv(
+                        self.LENGTH_LENGTH - len(block_length_bytes)
+                    )
+                    if not block_length_bytes_remainder:
+                        return result
```
ikalchev, line 746: How about we change the way we read the length header like `block_length_bytes = self.socket.recv(self.LENGTH_LENGTH, flags=socket.MSG_WAITALL)`, forcing the socket to get us LENGTH_LENGTH before returning? This would simplify the logic a lot in exchange for crappy performance in the case where a block ends just at `LENGTH_LENGTH - 1` (which should seldom happen).

bdraco: Looks like that has broad cross OS support.  Good call. I'll rework and then test for a bit

bdraco:
Testing...

pyhap/hap_server.py
bdraco: So far so good.  I'll keep testing over the rest of the day

bdraco:
Switched over 4 Home Assistant installs to the new code.  so far so good

bdraco:
Stable and very responsive with 347 devices linked between 4 hass installs

bdraco:
All good overnight as well

bdraco:
@ikalchev Any chance you have a spare moment to get another version bump?

ikalchev:
Yes I will do it tonight

ikalchev:
2.8.1 released. Let me know if something's wrong. I now see one test started failing, but I think it's the test itself that has issues. Will look more into it in the coming days.

bdraco:
@ikalchev Thank you.  Looks like its the camera test failing.  Since Home Assistant doesn't have HAP camera support yet if there is something wrong it likely won't get noticed.

PR #241, opened 2020-04-15
Add logging of exceptions in finish_request

whooo:
If a request fails with an exception other the socket timeout there
is no information of why, so add debug logging.

To give some background for this change I by accident built the cryptography package against a openssl fork that doesn't support ChaCha20Poly1305, which meant that my home-assistant server just closed all connections when getting requests from my phone without any information in the logs on why.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -893,6 +893,8 @@ class HAPServer(socketserver.ThreadingMixIn,
         except (OSError, socket.timeout) as e:
             self._handle_sock_timeout(client_address, e)
             logger.debug('Connection timeout')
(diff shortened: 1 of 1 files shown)


bdraco:
This changed caused a regression where homekit will loop forever trying to subscribe to events if there is a failure because it doesn't get the response it expects

bdraco:
I haven't figured out why it ends up looping though as I haven't been able to catch it with debug on

bdraco:
After digging into this I don't think this is the cause.  There is an existing issue where there is still data in the outgoing buffer when we upgrade to encrypted

PR #245, opened 2020-05-05
Fix exception on missing video fields

bdraco:
* Fix exception on missing video fields

Takes care of the `# TODO: Optionals, handle the case where they are missing`.  Sometimes streams would fail to start because a field would be missing in `VIDEO_TYPES['ATTRIBUTES']`.  Everything is starting consistently now with this change.

* Make logging consistent

Logging was calling `logging` instead of `logger` like the other modules.  This meant the debug log didn't show up in Home Asssistant.

### DIFF
--- a/pyhap/camera.py
+++ b/pyhap/camera.py
@@ -541,8 +541,10 @@ class Camera(Accessory):
         if success:
             self.streaming_status = STREAMING_STATUS['STREAMING']
         else:
-            logger.error('[%s] Faled to start/reconfigure stream, deleting session.',
-                          session_id)
+            logger.error(
+                '[%s] Failed to start/reconfigure stream, deleting session.',
+                session_id
+            )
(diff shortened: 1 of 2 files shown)


bdraco:
Also Home Assistant now has HomeKit camera support in the upcoming 0.110 🍾

ikalchev:
I suppose a release with the fix is needed. Anything else I should wait for before doing it?

bdraco:
> I suppose a release with the fix is needed. Anything else I should wait for before doing it?

Nothing else in the queue so a release would be appreciated. Thank you!

bdraco:
Well @ikalchev, looks like I was wrong there is one more thing I found when I was adding and removing test bridges for cameras https://github.com/ikalchev/HAP-python/pull/246

Thanks again for the reviews and releases. 🎉

PR #246, opened 2020-05-08
Fix race condition that causes pairing and unpairing failures

bdraco:
The mDNS announcement must not be updated until AFTER
the final pairing response is sent or homekit will
see that the accessory is already paired and assume
it should stop pairing as it assumes it lost the
race and another controller has paired.

This probably means safe mode can go away but I'm
not 100% sure on that.


bdraco:
With Home Assistant now supporting multiple bridges, I've noticed that paired failed about 20% of the time and finally tracked it down to pairing aborting when the new mdns announcement happens which causes pairing to cancel because it sees the bridge as already paired.

PR #252, opened 2020-05-14
Add support for accessory unavailability

bdraco:
Implements #194 

Example usage
https://github.com/bdraco/home-assistant/commit/a37b1fb0c4839f496aeea665d7f575ec66f9587c

Adds an exception guard to prevent the encrypted connection from being dropped and causing homekit to stall usually caused by the `getter_callback` throwing an unexpected exception.

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -74,6 +74,18 @@ class Accessory:
         """
         pass
 
+    @property
+    def available(self):
+        """Accessory is available.
+
+        If available is False, get_characteristics will return
+        SERVICE_COMMUNICATION_FAILURE for the accessory which will
+        show as unavailable.
+
+        Expected to be overridden.
+        """
+        return True
+
     def add_info_service(self):
         """Helper method to add the required `AccessoryInformation` service.
 
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -47,7 +47,6 @@ class Accessory:
         self.aid = aid
         self.display_name = display_name
         self.driver = driver
-        self.reachable = True
         self.services = []
         self.iid_manager = IIDManager()
 
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -626,16 +626,30 @@ class AccessoryDriver:
         :rtype: dict
         """
         chars = []
-        for id in char_ids:
-            aid, iid = (int(i) for i in id.split('.'))
-            rep = {HAP_REPR_AID: aid, HAP_REPR_IID: iid}
-            char = self.accessory.get_characteristic(aid, iid)
+        for aid_iid in char_ids:
+            aid, iid = (int(i) for i in aid_iid.split("."))
+            rep = {
+                HAP_REPR_AID: aid,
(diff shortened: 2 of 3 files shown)


bdraco:
Not sure what is going on with the coverage check but will look into it after the work day

bdraco:
4dea6879bf01f944402a4069ca2c4c7207f40428 is the first bad commit
commit 4dea6879bf01f944402a4069ca2c4c7207f40428
Author: Erik Larsson <who+github@cnackers.org>
Date:   Wed Apr 15 09:56:31 2020 +0200

    Add logging of exceptions in finish_request
    
    If a request fails with an exception other the socket timeout there
    is no information of why, so add debug logging.

:040000 040000 9103d69a6c3b73f841bc8a6586d78ed180c330dc 3c90b0febf5c0c5023df9168bed8fc889bf9b12b M	pyhap
bisect run success

bdraco:
Probably need to rebase this after the fix in https://github.com/ikalchev/HAP-python/pull/254

bdraco:
https://travis-ci.org/github/ikalchev/HAP-python/builds/687437366

Travis is clean now

pyhap/accessory.py
```diff
@@ -74,6 +74,18 @@ def _set_services(self):
         """
         pass
 
+    @property
+    def available(self):
```
nikolaykasyanov, line 78: This makes the `self.reachable = True` assignment in init obsolete.

pyhap/accessory_driver.py
```diff
@@ -626,16 +626,30 @@ def get_characteristics(self, char_ids):
...
-        for id in char_ids:
-            aid, iid = (int(i) for i in id.split('.'))
-            rep = {HAP_REPR_AID: aid, HAP_REPR_IID: iid}
-            char = self.accessory.get_characteristic(aid, iid)
+        for aid_iid in char_ids:
+            aid, iid = (int(i) for i in aid_iid.split("."))
+            rep = {
+                HAP_REPR_AID: aid,
+                HAP_REPR_IID: iid,
+                HAP_REPR_STATUS: SERVICE_COMMUNICATION_FAILURE,
```
nikolaykasyanov, line 634: I wonder if it'd make sense to use some other code to render the accessory completely unavailable from HomeKit's perspective (see [comment](https://github.com/homebridge/HAP-NodeJS/issues/499#issuecomment-335736097)).

pyhap/accessory.py
bdraco: I think its can simply go away since its not used anywhere in the code AFAICT

pyhap/accessory_driver.py
bdraco: That is exactly what is going on here (completely unavailable to homekit)

Screen Shot 2020-05-17 at 9 54 18 AM

nikolaykasyanov: Great, thanks for clarifying.

bdraco:
Travis is clean https://travis-ci.org/github/ikalchev/HAP-python/builds/690088137

PR #253, opened 2020-05-15
Fix random disconnect after upgrade to encrypted

bdraco:
After server restarts accessories would sometimes go
unavailable because after reconnecting and switching
to encrypted transport part of the unencrypted response
would not have been sent down the wire.  After the
upgrade to encrypted happens that response gets lost
which means the server starts sending encrypted data
when the hap client/controller is still expecting
to finish reading the unencrypted data.

The net result is that the controller/client has to
try again to get another connection to the server
where it does not get unluckly and hit this race
condition.

Before the change "Cleaning connection" was
very common in the log.  Its now rare.

Now that Apple has published more information,
we can improve conformance to HAP spec.

Content-Length is no longer sent on empty content
as it could lead to a loop on the client/controller side

The Connection, Date, and Server headers are not expected to be
sent according to the spec, and they are now suppressed.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -157,6 +157,7 @@ class HAPServerHandler(BaseHTTPRequestHandler):
         # it can be painfully slow and lead to lock up on the
         # client side as well as non-responsive devices
         self.protocol_version = 'HTTP/1.1'
+        self.status_code = None
         # Redirect separate handlers to the dispatch method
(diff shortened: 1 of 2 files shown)


bdraco:
Manual Testing steps:

Turn on airplane mode, turn off wifi, wait for accessories to go unavailable.

Turn back on wifi and airplane mode off. Watch the log.  Verify no double connects

ikalchev:
Great job!

PR #254, opened 2020-05-15
Ensure exception is still thrown in finish_request

bdraco:
The "Add logging of exceptions in finish_request" PR caused the exception to be suppressed instead of only adding logging.


bdraco:
Should fix the failing test

bdraco:
Travis is happy again 

https://travis-ci.org/github/ikalchev/HAP-python/builds/687287408

PR #256, opened 2020-05-16
Only stringify the characteristic uuid once

bdraco:
I've been working on making sure everything works with a RPi and a large number of accessories.  As part of of that I've been looking at performance bottlenecks.

I noticed that this was ~25% of the execution time when calling `get_accessories`.


bdraco:
Travis is clean https://travis-ci.org/github/ikalchev/HAP-python/builds/687796372

Issue #257, opened 2020-05-26
Cleaning connection for no apparent reason, resulting in No Response


SebasT87:
I have been breaking my head around an issue where my non-HomeKit devices published through Home Assistant are not responding every couple of minutes. My true-HomeKit devices are no problem though.

The problem starts when the connection gets cleaned (for no apparant reason), see log

> 2020-05-26 08:01:01 DEBUG (Thread-17) [pyhap.accessory_driver] Subscribed client ('192.168.1.25', 49795) to topic 2225605144.9
> 2020-05-26 08:01:01 DEBUG (Thread-17) [pyhap.accessory_driver] Subscribed client ('192.168.1.25', 49795) to topic 2225605144.10
> 2020-05-26 08:01:01 INFO (Thread-17) [pyhap.hap_server] 192.168.1.25 - "PUT /characteristics HTTP/1.1" 204 -
> 2020-05-26 08:03:49 DEBUG (Thread-17) [pyhap.hap_server] Cleaning connection to ('192.168.1.25', 49795)

After this nothing gets logged for some time (somtimes up to a couple of minutes) and the devices are not responding. Then all of a sudden it connects and the devices start responding again. After a couple of minutes this loop starts over.

> 2020-05-26 08:09:07 INFO (Thread-5) [pyhap.hap_server] Got connection with ('192.168.1.25', 49796).
> 2020-05-26 08:09:07 DEBUG (Thread-18) [pyhap.hap_server] Request POST from address '('192.168.1.25', 49796)' for path '/pair-verify'.
> 2020-05-26 08:09:07 DEBUG (Thread-18) [pyhap.hap_server] Pair verify [1/2].
> 2020-05-26 08:09:07 INFO (Thread-18) [pyhap.hap_server] 192.168.1.25 - "POST /pair-verify HTTP/1.1" 200 -
> 2020-05-26 08:09:07 DEBUG (Thread-18) [pyhap.hap_server] Request POST from address '('192.168.1.25', 49796)' for path '/pair-verify'.
> 2020-05-26 08:09:07 DEBUG (Thread-18) [pyhap.hap_server] Pair verify [2/2]
> 2020-05-26 08:09:07 DEBUG (Thread-18) [pyhap.hap_server] Pair verify with client '('192.168.1.25', 49796)' completed. Switching to encrypted transport.
[...]

PR #258, opened 2020-05-26
Split read/write encryption upgrade

bdraco:
#253 only fixed the write side of the issue.

Once the final response is sent data, can come
back over the wire before the reader is swapped
out to the encrypted reader.  This causes us to
loose the data and result in a disconnect.

Should solve #257

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -194,22 +194,29 @@ class HAPServerHandler(BaseHTTPRequestHandler):
             "pre_session_key": pre_session_key
         }
 
-    def _upgrade_to_encrypted(self):
+    def _upgrade_reader_to_encrypted(self):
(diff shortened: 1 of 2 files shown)


bdraco:
Travis is clean. https://travis-ci.org/github/ikalchev/HAP-python/builds/691509872
We don't have good coverage in this area, but thats going to be beyond the scope of this bugfix.

bdraco:
#253 fixed it for wifi for me

This was only happening for me on desktop machines connected via ethernet and apple tvs because the latency was low enough.

bdraco:
https://travis-ci.org/github/ikalchev/HAP-python/builds/691517095

bdraco:
No more cleaning connections from 3 ethernet connected macs

ikalchev:
I was curious, how did you catch that?

PR #255, opened 2020-05-15
Various code cleanups, add py37/38 to travis

bdraco:
This is a work in progress and will conflict with https://github.com/ikalchev/HAP-python/pull/252 so I'll rework it after that

PR #259, opened 2020-05-26
Fix pairing after py37/py38 cleanup

bdraco:
I missed removing a `, False)` in the original PR. 🙈


bdraco:
@ikalchev I caused a regression in #255.  This should definitely go in before release.

bdraco:
https://travis-ci.org/github/ikalchev/HAP-python/builds/691517095

bdraco:
Reverified ok on another test system

PR #269, opened 2020-06-29
✨ Support multiple accessories with same name

jerr0328:
I was running into an issue trying to deploy the same code for an accessory on two different RPis implemented with this library, with different hostnames. As soon as the second one was on, neither accessory would work anymore, they were both getting some corrupted/bad requests. They both had the same display_name, so I changed one and suddenly everything was working correctly. But this was weird, as I thought the autodiscovery features should account for this. That led me to this SO answer: https://stackoverflow.com/a/53867271/375530

Comparing what other HomeKit accessories do, it seems they append some parts of their MAC address to the mDNS name, giving some uniqueness if there are multiple bridges on the same network, for instance. Since the accessory I wrote has multiple physical copies, appending the MAC address would solve the issue. I have tested this by running this code on my Pi and seeing the new name get picked up correctly without clashing.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -56,6 +56,7 @@ CHAR_STAT_OK = 0
 SERVICE_COMMUNICATION_FAILURE = -70402
 SERVICE_CALLBACK = 0
 SERVICE_CALLBACK_DATA = 1
+HAP_SERVICE_TYPE = '_hap._tcp.local.'
 
 
 def callback(func):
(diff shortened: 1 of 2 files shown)


pyhap/accessory_driver.py
```diff
@@ -86,9 +87,15 @@ def __init__(self, accessory, state):
...
+        name = '{} {}.{}'.format(
+            self.accessory.display_name,
+            self.state.mac[-8:].replace(':', ''),
+            HAP_SERVICE_TYPE
+        )
         super().__init__(
-            '_hap._tcp.local.',
-            name=self.accessory.display_name + '._hap._tcp.local.',
+            HAP_SERVICE_TYPE,
+            name=name,
```
bdraco, line 98: Will this break existing pairings that are looking for the old name ?

jerr0328: When I tested it, I had already paired an accessory and then changed the name and it remained paired in Home. I'm guessing that because the data (the ID?) in the advertisement payload isn't changed, HomeKit knows it's the same device and keeps it.

bdraco: I'll test this later today to make sure its not going to cause anything fallout with Home Assistant

bdraco: It looks like we limit the name length in Home Assistant to 25 chars 
https://github.com/home-assistant/core/pull/15600

That should ensure we don't hit the mdns label max length of 63 chars

jerr0328:
Please let me know if there's anything else needed to get this merged :)

PR #270, opened 2020-07-03
Improve loop handling

balloob:
This improves loop handling for HAP-Python:
 - Do not change the default executor of passed in loops
 - Do not shut down loops/executor pools that were passed in

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -70,7 +70,7 @@ def is_callback(func):
 
 
 def iscoro(func):
-    """Check if the function is a coroutine or if the function is a ``functools.patial``,
+    """Check if the function is a coroutine or if the function is a ``functools.partial``,
     check the wrapped function for the same.
     """
     if isinstance(func, functools.partial):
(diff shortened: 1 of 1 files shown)


bdraco:
```
            if threading.current_thread() is threading.main_thread():
                logger.debug('Setting child watcher')
                watcher = asyncio.SafeChildWatcher()
                watcher.attach_loop(self.loop)
                asyncio.set_child_watcher(watcher)
```

I think we need a way to call `_do_start` without the above in `start`

balloob:
Done, renamed `_do_start` to `start_service`

bdraco:
Everything appears to work as expected in the linked PR (testing using the existing protected methods)

bdraco:
travis is happy
https://travis-ci.org/github/ikalchev/HAP-python/builds/704797505

pyhap/accessory_driver.py
```diff
@@ -210,7 +215,7 @@ def __init__(self, *, address=None, port=51234,
         self.topics = {}  # topic: set of (address, port) of subscribed clients
         self.topic_lock = threading.Lock()  # for exclusive access to the topics
         self.loader = loader or Loader()
-        self.aio_stop_event = asyncio.Event(loop=self.loop)
+        self.aio_stop_event = asyncio.Event()
```
balloob, line 218: This change is breaking Python 3.5 support. It's not necessary, just a drive by cleanup from when I saw warnings printed during tests. I will revert it.

bdraco:
Verified this doesn't regress busy home example

bdraco:
@ikalchev If this one looks good, could we get a new release? 

Thank you.

ikalchev:
Change is good, will release in the evening.

Best,
Ivan

bdraco:
Thanks Ivan 👍

PR #274, opened 2020-07-20
Use SimpleQueue when available (python 3.7 and later)

bdraco:
`SimpleQueue` avoids a lot of the overhead of `Queue`.  Since the current implementation doesn't do anything advanced with the queue, we can benefit from the increased performance


bdraco:
Tested on python 3.7 and python 3.8

The travis run should tell us if 3.5/3.6 is an issue

PR #297, opened 2020-12-15
Properly handle snapshots if camera added to bridge.

bzzeke:
The `handle_resource` method from `HAPServer` class does not support situation when camera added to bridge and fails with `Got a request for snapshot, but the Accessory does not define a "get_snapshot" method` error. This is because it tries to check if bridge itself has the `get_snapshot` method.

This pull request fixes this problem.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -26,7 +26,7 @@ import ed25519
 
 import pyhap.tlv as tlv
 from pyhap.util import long_to_bytes
-from pyhap.const import __version__
+from pyhap.const import __version__, CATEGORY_BRIDGE
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


pyhap/hap_server.py
```diff
@@ -664,13 +664,23 @@ def _handle_remove_pairing(self, tlv_objects):
 
     def handle_resource(self):
         """Get a snapshot from the camera."""
-        if not hasattr(self.accessory_handler.accessory, 'get_snapshot'):
+
+        data_len = int(self.headers["Content-Length"])
+        request_body = self.rfile.read(data_len)
+        data = json.loads(request_body.decode("utf-8"))
+
+        if self.accessory_handler.accessory.category == CATEGORY_BRIDGE:
```
bdraco, line 672: Can we check for `STANDALONE_AID` instead?

bzzeke: Just tried, but `self.accessory_handler.accessory.aid` returns `1` both for bridge and standalone accessory

bzzeke: I've created new pull request which merges correctly with 3.2.0.
https://github.com/ikalchev/HAP-python/pull/306

bdraco:
I think this can be closed in favor of https://github.com/ikalchev/HAP-python/pull/306

bdraco (changes requested):
I think this can be closed in favor of https://github.com/ikalchev/HAP-python/pull/306

PR #299, opened 2021-01-05
Timeout client sockets after 300 seconds 

bdraco:
This is an adjusted version of #222 that ensures that the client knows the connection is closed by sending a 500 response when it the socket times out.  When testing #222 I was seeing `No Response` on devices when the socket timed out because the client thought it was still connected.

### DIFF
--- a/pyhap/hap_server.py
+++ b/pyhap/hap_server.py
@@ -279,6 +279,9 @@ class HAPServerHandler(BaseHTTPRequestHandler):
             self.send_response_with_status(401, HAP_SERVER_STATUS.INSUFFICIENT_PRIVILEGES)
         except TimeoutException:
             self.send_response_with_status(500, HAP_SERVER_STATUS.OPERATION_TIMED_OUT)
+        except socket.timeout:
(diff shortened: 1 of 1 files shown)


bdraco:
@vladimirlagunov Does this solve the issue for you still?  I can confirm that the `No response` behavior I originally saw with #222 doesn't appear with the additional commit.

vladimirlagunov:
@bdraco I use the library as a part of HomeAssistant. I've updated HomeAssistant twice in my life, both times brought me several hours of sufferings with debugging and recreating everything from scratch. On the other hand, I use it only to control a simple light switch in a private network, so the current functionality is far enough for me. I promised myself to never ever update home assistant and any of its dependencies. Sorry, I'm keeping my own patch and won't check any other approach.

bdraco:
Thanks for responding.  I've been running this for a few days now without any issues and I can confirm no extra threads are left around.

I'll mark it as ready for review.  Hopefully this will help others who have threads building up

bdraco:
Probably not needed with https://github.com/ikalchev/HAP-python/pull/301

PR #301, opened 2021-01-28
Convert to using asyncio for the http server

bdraco:
All the extra threads are gone, and its using an `asyncio.Protocol`, and `h11` is used to handle all the http protocol.

`hap_server.py` was broken up into `hap_crypto.py`, `hap_handler.py`, and `hap_protocol.py` to make the code more testable now that all the I/O is separated out.  Now with coverage on processing the encrypted buffer.

Seems to 100% solve unexplained dropped connections (no more lost connections in the log).

If travis is still working https://travis-ci.org/github/ikalchev/HAP-python/builds/756828950


bdraco:
Need to address a bug introduced from the squash resolution

bdraco:
If travis is still working
https://travis-ci.org/github/ikalchev/HAP-python/builds/756526026

bdraco:
@ikalchev Can you take a look?

ikalchev:
That's wonderful 🎉! Will definitely go over all of these over the weekend, for sure.

bdraco:
Last push was to add additional coverage and fix python 3.5

bdraco:
Also fixed a subtle bug in the barcode/url https://github.com/ikalchev/HAP-python/pull/303

bdraco:
Will keep pushing tests as I have time to write them.  Not expecting any of the non-test code to change at this point.

bdraco:
travis timed out access github.  I'm going to push it again just to see if codecov will update since there is more coverage now.

bdraco:
Pushing again to rebase to dev (no changes)

pyhap/hap_server.py
```diff
@@ -895,93 +37,36 @@ def create_hap_event(cls, bytesdata):
...
+
+    async def async_start(self, loop):
+        """Start the http-hap server."""
+        self.server = await loop.create_server(
+            lambda: HAPServerProtocol(loop, self.connections, self.accessory_handler),
+            self._addr_port[0],
+            self._addr_port[1],
+        )
+
+    def async_stop(self):
```
ikalchev, line 63: I am a bit confused about the method name and not being async. Can you elaborate? Saw something similar in the accessory driver as well.

bdraco: The method must be called in the event loop and not from the executor as it is not thread safe.

Aka it's only safe to call in async context, but since nothing is being awaited it doesn't need to be a coroutine. 

Usually in home assistant we decorate these with @callback

bdraco: Looks like we have `@callback`.  I'll add that now

ikalchev: Ah ok, makes sense now.

bdraco:
Thank you!  Would you mind pushing a release so I can update Home Assistant?

ikalchev:
Definitely! First thing after my coffee tomorrow :) Great work!

ikalchev:
@bdraco 3.2.0 should be in with all the added goodness :)

bdraco:
TYVM!

bdraco:
Retested and all working as expected

PR #302, opened 2021-01-28
Make category an instance attribute

bdraco:
Since it was a class attribute, the category would
be presented to homekit incorrectly when there
were multiple bridged devices


bdraco:
Actually I think Home Assistant is just doing it wrong.  This is fine as-is

PR #307, opened 2021-02-06
Implement list pairings

bdraco:
Pairing would sometimes fail because list pairings
was not implemented and an exception would be generated.

The state file could get corrupted if something went
wrong with the write. We now write a temp file and
move it in place after its fully written.

Ensure writing the state happens outside
of the event loop.

Also fixes connection stall on camera image failure.

Project test coverage up to 82.65% 

Fixes

```
2021-02-02 08:33:15 ERROR (MainThread) [pyhap.hap_handler] Failed to process request for: /pairings
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/pyhap/hap_handler.py", line 207, in dispatch
    getattr(self, self.HANDLERS[self.command][path])()
  File "/usr/local/lib/python3.8/site-packages/pyhap/hap_handler.py", line 608, in handle_pairings
    raise ValueError(
ValueError: Unknown pairing request type of 5 during pair verify
```

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -20,6 +20,7 @@ import base64
 from concurrent.futures import ThreadPoolExecutor
 import functools
 import hashlib
+import tempfile
 import json
 import logging
 import os
@@ -48,6 +49,7 @@ from pyhap.hsrp import Server as SrpServer
 from pyhap.loader import Loader
 from pyhap.params import get_srp_context
 from pyhap.state import State
+from .util import callback
 
 logger = logging.getLogger(__name__)
 
@@ -305,9 +307,11 @@ class AccessoryDriver:
             self.accessory.setup_message()
 
         # Start the accessory so it can do stuff.
-        logger.debug("Starting accessory.")
+        logger.debug("Starting accessory %s", self.accessory.display_name)
         self.add_job(self.accessory.run)
-        logger.debug("AccessoryDriver started successfully")
+        logger.debug(
+            "AccessoryDriver for %s started successfully", self.accessory.display_name
+        )
 
     def stop(self):
         """Method to stop pyhap."""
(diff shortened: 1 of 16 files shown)


pyhap/hap_handler.py
```diff
@@ -596,60 +600,93 @@ def handle_set_characteristics(self):
...
-        self.send_response(200)
-        self.send_header("Content-Type", self.PAIRING_RESPONSE_TYPE)
-        self.end_response(data)
-
-        # Avoid updating the announcement until
-        # after the response is sent as homekit will
-        # drop the connection and fail to pair if it
-        # sees the accessory is now paired as it doesn't
-        # know that it was the one doing the pairing.
-        self.accessory_handler.finish_pair()
```
bdraco, line 635: The mdns announcement was being updated every time a new appletv/hub was added.  We only need to set the `sf` once when we finish `_pairing_five`

bdraco:
@ikalchev If this looks good, can you bump another release?

pyhap/accessory_driver.py
```diff
@@ -524,14 +551,12 @@ def pair(self, client_uuid, client_public):
...
-        # let the accessory call config_changed, which will persist and update mDNS?
-        # See also unpair.
         logger.info("Paired with %s.", client_uuid)
         self.state.add_paired_client(client_uuid, client_public)
-        self.persist()
+        self.async_persist()
         return True
 
+    @callback
     def unpair(self, client_uuid):
```
bdraco, line 560: I didn't rename this to `async_unpair` since it may be overridden in a child class

pyhap/accessory_driver.py
```diff
@@ -510,6 +536,7 @@ def load(self):
         with open(self.persist_file, "r") as file_handle:
             self.encoder.load_into(file_handle, self.state)
 
+    @callback
     def pair(self, client_uuid, client_public):
```
bdraco, line 540: I didn't rename this to `async_pair` since it may be overridden in a child class

PR #306, opened 2021-02-01
Properly handle snapshots if camera added to bridge

bzzeke:
[Updated pull request to work correctly with 3.2.0]

The handle_resource method from HAPServer class does not support situation when camera added to bridge and fails with Got a request for snapshot, but the Accessory does not define a "get_snapshot" method error. This is because it tries to check if bridge itself has the get_snapshot method.

This pull request fixes this problem.

PR #308, opened 2021-02-10
Speed up new HAP event subscribe on new connections

bdraco:
Avoids the linear search of objects to
find the iid. When there are a lot of
characteristics and accessories on a
bridge, the reconnect could cause the
bridge to go unavailable on slower systems (RPi3)
while it was re-subscribing to events.

We now use a dict lookup to find object
from the iid

### DIFF
--- a/pyhap/iid_manager.py
+++ b/pyhap/iid_manager.py
@@ -9,8 +9,9 @@ class IIDManager:
 
     def __init__(self):
         """Initialize an empty instance."""
-        self.iids = {}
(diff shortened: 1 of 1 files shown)


bdraco:
@ikalchev  

If you have a moment the following should be ready for review & release.

https://github.com/ikalchev/HAP-python/pull/306
https://github.com/ikalchev/HAP-python/pull/307
https://github.com/ikalchev/HAP-python/pull/308

PR #312, opened 2021-02-12
Switch codecov to use gh actions and drop travis

bdraco:
May require installing https://github.com/apps/codecov/installations/new

Drops travis since its going away


ikalchev:
Thanks for that!

PR #324, opened 2021-03-03
Cleanup subscriptions on client disconnect

bdraco:
- Solves a memory leak with topics
- Increase test coverage


ikalchev:
Can't believe we catch that now.
Great work!

PR #328, opened 2021-03-05
Support python 3.10

### LINKED ISSUE #304: Do not pass `loop` to asyncio.Event
Passing the loop argument to constructors is deprecated in Python 3.8 and will be removed in Python 3.10. You can trigger this as an error now by running the test suite with error on warning:

```
py.test tests/ -Werror -x
```

Making this change would require to remove all passing of / creation of loop in the accessory.

I tried to do this quickly but it means that the accessory needs to be instantiated from inside the event loop, which is not happening inside the tests.

bdraco:
Fixes #304


bdraco:
The removed coverage is restored in https://github.com/ikalchev/HAP-python/pull/320 (was adjusted to merge cleanly)

PR #332, opened 2021-03-06
Complete hap_protocol coverage

bdraco:
- Improves handling of client closing connection cleanly


bdraco:
@ikalchev That should be it for a while.  At some point I'd like to add full coverage for pair and pair verify but thats for another day.

Tested and should be good to go for release.

PR #337, opened 2021-03-23
Ensure HTTP 200 status is sent when get_characteristics has no failures

bdraco:
Previously we always sent a multi-status of 207 even when
there were no failures. HomeKit expects a HTTP 200 response
when there are no failures, and for status to be not
be set in the 100% success state.

This change **drastically** reduces the amount of polling that
the iOS device does because it no longer schedules a retry
based on the unexpected 207 which translates into a meaningful
improvement in battery life.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -51,12 +51,11 @@ from pyhap.loader import Loader
 from pyhap.params import get_srp_context
 from pyhap.state import State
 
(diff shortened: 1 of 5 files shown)


bdraco:
@ikalchev Can we get a release with this one when you have a free moment?

ikalchev:
Sure, will merge and release tomorrow.

bdraco:
Thank you 🙏🏻

PR #342, opened 2021-05-06
Use zeroconf async registration functions

bdraco:
Creating an `AccessoryDriver` now accepts an `AsyncZeroconf` instance via `async_zeroconf_instance` instead of a `Zeroconf` instance via `zeroconf_instance`.  No changes are needed if a zeroconf instances was not being passed in.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -28,7 +28,8 @@ import sys
 import tempfile
 import threading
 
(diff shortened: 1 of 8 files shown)


bdraco:
@ikalchev When you get a chance, can you push a release?  I'd like to get this in this month's Home Assistant release. Thanks!

PR #345, opened 2021-05-25
Reduce payload sizes by adding support for short UUIDs and compact json

bdraco:
`/accessories` payloads sizes are now reduced by ~40%+ with:

- Apple defined uuids should be reduced to their short form to reduce the payloads

- `description` (optional) is only included in the payload if it has been customized

- compact json

- fixes events not being utf-8 encoded

This solved another case of iOS getting stuck updating when the bridge has many devices


bdraco:
This does quite a bit to reduce payload sizes and improve performance since the encrypted payloads are much smaller.

bdraco:
@ikalchev Can we get a release with this for this week's Home Assistant beta?

Thanks!

PR #348, opened 2021-06-09
Handle additional cases of invalid hostnames

bdraco:
Some mDNS implementations were crashing when seeing the `SRV` records with a space which was translating to `HASS\032Bridge\032._hap._tcp.local.`


bdraco:
Once I clean this up and add the missing coverage, need to do another PR to prep for the zeroconf namespace change from `zeroconf.asyncio -> zeroconf.aio`

We can do a condition import for the transition

```python3
try:
   import AsyncZeroconf from zeroconf.aio
except ImportError:
   import AsyncZeroconf from zeroconf.asyncio
```

PR #349, opened 2021-06-18
Add support for zeroconf 0.32.0+

bdraco:


bdraco:
@ikalchev When you get a moment, can you push a new release with this one and https://github.com/ikalchev/HAP-python/pull/348 ?

bdraco:
We are going to try to rename this back now that `import asyncio` isn't needed in `zeroconf/__init__.py`

https://github.com/jstasiak/python-zeroconf/issues/883

bdraco:
It looks like we are going to be able to avoid the namespace change

https://github.com/jstasiak/python-zeroconf/pull/885

PR #355, opened 2021-07-17
Switch to using cryptography ed25519 

bdraco:
- The ed25519 package is no longer recommended by the maintainer https://github.com/warner/python-ed25519#not-recommended-for-new-applications-use-pynacl-instead

- Since we are already using cryptography, it makes sense to use that instead of pynacl

- In testing this was ~2.8x faster which speeds up the transition from cellular to wifi when coming home

### DIFF
--- a/pyhap/encoder.py
+++ b/pyhap/encoder.py
@@ -6,7 +6,8 @@ it can work properly after a restart.
 import json
 import uuid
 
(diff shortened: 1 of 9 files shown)


ikalchev:
Awesome work as always! 🚀

PR #356, opened 2021-07-17
Switch to using cryptography x25519

bdraco:
- Drop curve25519-donna as its unmaintained

Requires https://github.com/ikalchev/HAP-python/pull/355


bdraco:
Thanks!

PR #360, opened 2021-07-24
Reduce event overhead

### LINKED ISSUE #359: Events are sent for characteristics that have not changed
When `client_update_value` or `set_value` is called an event is generated even when the internal state hasn't changed. For cameras motion sensors this can result in 100x of events per hour per camera

### LINKED ISSUE #364: Sending an event after an unsubscribe has received a response can cause iOS to disconnect
-> client sends b'PUT /characteristics HTTP/1.1\r\nHost: HASS\\032Bridge\\032OS95\\0327D4B7B._hap._tcp.local\r\nContent-Length: 96\r\nContent-Type: application/hap+json\r\n\r\n{"characteristics":[{"aid":969756015,"iid":9,"ev":false},
{"aid":969756015,"iid":10,"ev":false}]}'

Event is queued for 969756015.9

<- b'HTTP/1.1 204 No Content\r\n\r\n'

Event is sent

Client disconnects because it was not expecting the event

bdraco:
- Only send an event if the characteristic has changed value

- Coalesce events that happen within the same 500ms window except
  for Button Events and Programmable Switch Events which are delivered
  as soon as possible

- Ensure Programmable Switch Events always serialize to null
  to handle the special case in the spec (except for events).

- MAJOR: Sending an event after the client has been unsubscribed may cause
  iOS to disconnect the accessory and reconnect which makes everything
  go unavailable while the connection is set back up. We now check to
  make sure that the client still wants the event right before sending it
  to avoid the disconnect.

Fixes #359 Fixes #364

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10.0-beta.4"]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v1
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -283,7 +283,7 @@ class Accessory:
 
     # Driver
 
-    def publish(self, value, sender, sender_client_addr=None):
+    def publish(self, value, sender, sender_client_addr=None, immediate=False):
         """Append AID and IID of the sender and forward it to the driver.
 
         Characteristics call this method to send updates.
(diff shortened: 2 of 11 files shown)


bdraco:
I've only been able to test remotely, but this looks much better as the event stream isn't a flood now.  When I get home tomorrow I'll test locally before marking ready.

bdraco:
Looks good locally. Also fixes a disconnect when the iOS device gets overloaded with events so I'd like to get it into Home Assistant before Tuesday so it can make beta.

bdraco:
Actually I should add tests for programable switch events

bdraco:
All looks good. I'm going to go find a color light and test with that as well

PR #365, opened 2021-07-25
Protocol 1.1: Upgrade protocol version (HAP-python 4.0.0)

bdraco:
TODO:

- [x] Implement timed writes https://github.com/ikalchev/HAP-python/pull/366

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -2,21 +2,27 @@
 import itertools
 import logging
 
+from uuid import UUID
 from pyhap import SUPPORT_QR_CODE, util
 from pyhap.const import (
     CATEGORY_BRIDGE,
     CATEGORY_OTHER,
     HAP_REPR_AID,
     HAP_REPR_IID,
+    HAP_PROTOCOL_VERSION,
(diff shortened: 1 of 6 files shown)


bdraco:
Looks like we are almost there but we have to implement the `/prepare` endpoint

ikalchev:
Will wait for a final nod from you before merging this and releasing

ikalchev:
Also I think some tests fail, mind if you take look if you do any final touches here?

bdraco:
On it

ikalchev:
Thanks! I suspect sloppy merging on my part, so apologies.

bdraco:
One of the PRs changed the behavior of the other.

All fixed in

https://github.com/ikalchev/HAP-python/pull/369

bdraco:
Ideally what is merged now can be released as 3.5.3 or 3.6.0 and then merge this and release at 4.0.0 so we can do a bump in Home Assistant for 3.x and then in next month 4.x

bdraco:
rebased to make sure this is clean against dev

PR #367, opened 2021-07-27
Decrease snapshot timeout to avoid being disconnected

bdraco:
- iOS will terminate the connection if it does not respond within
  10 seconds, so we only allow 9 seconds to avoid this.

- We no longer need to pause reading when taking a snapshot
  as we have a working timeout

### DIFF
--- a/pyhap/hap_protocol.py
+++ b/pyhap/hap_protocol.py
@@ -113,7 +113,6 @@ class HAPServerProtocol(asyncio.Protocol):
             + self.conn.send(h11.Data(data=response.body))
             + self.conn.send(h11.EndOfMessage())
         )
-        self.transport.resume_reading()
 
     def finish_and_close(self):
         """Cleanly finish and close the connection."""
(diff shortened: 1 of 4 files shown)


pyhap/hap_protocol.py
```diff
@@ -195,7 +195,13 @@ def _process_one_event(self) -> bool:
 
         if isinstance(event, h11.EndOfMessage):
             self.transport.pause_reading()
+            time_before = time.time()
             response = self.handler.dispatch(self.request, bytes(self.request_body))
+            logging.debug(
+                "Handling request: %s took %s seconds",
+                self.request,
+                time.time() - time_before,
+            )
```
bdraco, line 204:
```suggestion
```

pyhap/hap_protocol.py
```diff
@@ -195,7 +195,13 @@ def _process_one_event(self) -> bool:
 
         if isinstance(event, h11.EndOfMessage):
             self.transport.pause_reading()
+            time_before = time.time()
```
bdraco, line 198:
```suggestion
```

PR #374, opened 2021-08-10
Add color temperature lookup table

bdraco:
Thanks once again to HAP-NodeJS

Supports https://github.com/home-assistant/core/issues/54107


bdraco:
Unfortunately this seems like the colors still don't line up.

PR #375, opened 2021-08-17
Add Accessory Start Method

ckskate:
Adds support for a `start()` method/coroutine on accessories.

I wanted to do some async startup for an accessory, and found that a start method was pretty clean.

### DIFF
--- a/pyhap/accessory.py
+++ b/pyhap/accessory.py
@@ -293,6 +293,12 @@ class Accessory:
         Can be overridden with a normal or async method.
         """
 
+    async def start(self):
+        """Called to do any startup an accessory requires
+
+        Can be overridden with a normal or async method.
+        """
+
     async def stop(self):
         """Called when the Accessory should stop what is doing and clean up any resources.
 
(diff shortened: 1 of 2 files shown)


bdraco:
We have an existing `run` method that does essentially the same thing. Will that work for your use case?

https://github.com/ikalchev/HAP-python/blob/dev/pyhap/accessory_driver.py#L398

ckskate:
Unfortunately, I don't think so. I'm using the run method currently with the `runAtInterval` decorator to perform repeated updates, but I also need to do some one time async start-up connection stuff.

bdraco:
Maybe try creating a new function decorated with runatinterval and use run as your startup function

ckskate:
oh yeah, that's a good suggestion! I think I associated that decorator with the run method in my head bc of the sample code in the readme 😅. I believe I was also having issues with the driver's run loop not being the main event loop, which is why I included the `asyncio.set_event_loop(loop)` line. I'd be able to post the exact issue later, but I'm wondering if you have any suggestions/would that be a valid change?

bdraco:
> set_event_loop

When you create AccessoryDriver, pass the loop as an argument.

Example here https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit/__init__.py#L517

ckskate:
thank you for all the help!

PR #376, opened 2021-08-22
Increment the config version when the accessories change

bdraco:
- We now hash a sorted version of `/accessories` at start
  so we can tell when it changes and increment the config version

- Incrementing the config version tells iOS that it needs to
  evict its cache of /accessories to prevent invalid state

- Fixes accessories showing non-responsive or missing services
  because their services have changed and iOS is unaware because its
  using the old version of `/accessories`

### DIFF
--- a/pyhap/state.py
+++ b/pyhap/state.py
@@ -2,7 +2,12 @@
 from cryptography.hazmat.primitives.asymmetric import ed25519
 
 from pyhap import util
-from pyhap.const import CLIENT_PROP_PERMS, DEFAULT_CONFIG_VERSION, DEFAULT_PORT
+from pyhap.const import (
+    CLIENT_PROP_PERMS,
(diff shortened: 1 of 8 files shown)


pyhap/state.py
```diff
@@ -69,3 +71,13 @@ def remove_paired_client(self, client_uuid):
...
             self.client_properties.clear()
+
+    def set_accessories_hash(self, accessories_hash):
+        """Set the accessories hash and increment the config version if needed."""
+        if self.accessories_hash == accessories_hash:
+            return False
+        self.accessories_hash = accessories_hash
+        self.config_version += 1
+        if self.config_version > MAX_CONFIG_VERSION:
+            self.config_version = 1
```
bdraco, line 82:
```suggestion
        self.config_version = min(MAX_CONFIG_VERSION, self.config_version + 1)
```

PR #378, opened 2021-08-23
Handle additional cases of invalid values

bdraco:
- Ensure strings are never longer than the current value of
  maxLen. Previously we would increase maxLen when it was too
  short, but since the iOS device will not see the maxLen change
  until the config version has been incremented, this would result
  in it seeing it as an invalid value and disconnect

- Ensure any set numeric value observes minStep. If minStep was 2, it
  was possible to set a value that was invalid and cause a disconnect

### DIFF
--- a/pyhap/characteristic.py
+++ b/pyhap/characteristic.py
@@ -103,6 +103,15 @@ class CharacteristicError(Exception):
     """Generic exception class for characteristic errors."""
 
 
+def _validate_properties(properties):
+    """Throw an exception on invalid properties."""
+    if (
+        HAP_REPR_MAX_LEN in properties
+        and properties[HAP_REPR_MAX_LEN] > ABSOLUTE_MAX_LENGTH
+    ):
+        raise ValueError(f"{HAP_REPR_MAX_LEN} may not exceed {ABSOLUTE_MAX_LENGTH}")
+
+
 class Characteristic:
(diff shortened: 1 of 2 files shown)


pyhap/characteristic.py
```diff
@@ -184,7 +187,11 @@ def to_valid_value(self, value):
                 logger.error(error_msg)
                 raise ValueError(error_msg)
         elif self.properties[PROP_FORMAT] == HAP_FORMAT_STRING:
-            value = str(value)[:256]
+            max_length = min(
+                ABSOLUTE_MAX_LENGTH,
```
bdraco, line 191: It might make more sense to throw a value error in `override_properties` if they try to set it to something larger than `ABSOLUTE_MAX_LENGTH`

pyhap/characteristic.py
```diff
@@ -322,8 +332,12 @@ def to_HAP(self):
                     self.properties[PROP_VALID_VALUES].values()
                 )
         elif self.properties[PROP_FORMAT] == HAP_FORMAT_STRING:
-            if len(value) > 64:
-                hap_rep[HAP_REPR_MAX_LEN] = min(len(value), 256)
+            max_length = min(
```
bdraco, line 335: https://github.com/ikalchev/HAP-python/pull/378/files#r694203262

PR #393, opened 2021-12-02
Fix pylint errors in CI

bdraco:


bdraco:
Thanks @ikalchev would you mind pushing out a release so we can close out https://github.com/home-assistant/core/pull/60220 ?

PR #412, opened 2022-06-13
Add support for orjson

bdraco:
- orjson is now used. As Home Assistant has [added
  support](https://github.com/home-assistant/core/pull/73849) for it, it will automatically be used when HAP-python
  is used with HA

- python's built in json is much slower and this improves the reconnect time when switching from out of home to home, and helps reduce the risk we will hit the lower timeouts in iOS 16 and have the bridge go unavailable on slower platforms


pyhap/util.py
```diff
@@ -13,6 +12,15 @@
...
 
+try:
+    from .orjson import (  # noqa: F401 # pylint: disable=unused-import
+        to_hap_json,
+        to_sorted_hap_json,
+        from_hap_json,
+    )
+except ImportError:
+    from .json import to_hap_json, to_sorted_hap_json, from_hap_json  # noqa: F401
+
```
ikalchev, line 23: Why don't you have a single `json.py` and, inside, try to import `orjson`:

```
try:
   import orjson
   def to_hap_....
except ImportError:
   import json
   def to_hap...
```

In that way all the feature switch logic is contained in the module providing json serialisation, i.e. dependent modules will not need to do the switch themselves.

bdraco: Thought about this one a bit and reloaded orjson has wider wheel availability and the same required build chain as cryptography.  I think it makes more sense to add it as a dep instead and eliminate the complexity of supporting both.  That's going to matter quite a bit anyways if iOS 16 actually ships with the lower timeouts as RPi3 users with large bridges are likely to see more unavailability without it anyways

PR #416, opened 2022-06-23
Add bump2version support

bdraco:


bdraco:
This one needs the pylint fixes in https://github.com/ikalchev/HAP-python/pull/413

setup.cfg
```diff
@@ -1,3 +1,13 @@
+[bumpversion]
+current_version = 4.4.0
```
bdraco, line 2:
```suggestion
current_version = 4.5.0
```

PR #417, opened 2022-06-23
Increase minimum python version to 3.7

bdraco:


bdraco:
This one needs the pylint fixes in https://github.com/ikalchev/HAP-python/pull/413

PR #423, opened 2022-10-08
Add support for a custom iid manager

bdraco:
Currently we rely on creation order to keep iids the same but this is not reliable since services and chars can change. To solve this we need a custom iid manager that tracks iid between restarts to ensure that iids remain stable for the lifetime of the accessory.

Needed to fix (only a small part) https://github.com/home-assistant/core/issues/78092 via https://github.com/home-assistant/core/pull/79913


tillsteinbach:
How can I set the unique_id for the AccessoryInformation Service when using the "set_info_service" method?

PR #428, opened 2022-11-29
Add necessary support for Adaptive Lightning

### LINKED ISSUE #284: Adaptive lighting support
(For those out of the loop: adaptive lighting is a new HomeKit feature pretty much similar to what Circadian Lighting offers, or so it sounds)
As far as I understood, the new Adaptive Lighting feature in HomeKit on iOS 14 is something that needs to be supported in the bridge.
Hue has announced support for it later this year, IKEA hasn’t officially pronounced themselves about it. Doing it from HAP-python would bring support for all the supported lights at the moment.

If I understand the stack correctly this should be possible. Also, for anyone who can offer me some pointers, I’d like to look into the pieces of code that would enable something like this.

Btw, I haven’t found any documentation about Adaptive Lighting so I’m only working from assumptions on how it all works.

vitkabele:
Add necessary characteristics to enable implementation of Adaptive Lightning. Also provide a demo light bulb implementation to demonstrate the basic usage of the newly added characteristics.

This PR closes #284

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,4 @@ pip-selfcheck.json
 
 # HAP-python-generated files
 accessory.pickle
-accessory.state
+/*.state
--- a/adaptive_lightbulb.py
+++ b/adaptive_lightbulb.py
@@ -1,0 +1,93 @@
+"""This virtual lightbulb implements the bare minimum needed for HomeKit
+    controller to recognize it as having AdaptiveLightning
+"""
+import logging
+import signal
+import random
+import tlv8
+import base64
+
+from pyhap.accessory import Accessory
+from pyhap.accessory_driver import AccessoryDriver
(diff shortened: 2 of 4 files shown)


niorko:
When it will be integrated into HA? :) I am looking forward to it. Thanks for the work.

vitkabele:
> When it will be integrated into HA? :) I am looking forward to it. Thanks for the work.

Not sure about that. I'm currently trying to figure out how exactly the integration should look. Please see related discussion at [HA forum](https://community.home-assistant.io/t/adaptive-lighting-support-for-homekit-ios-14/227225/9?u=vitkabele). I'll be happy for any suggestion.

Harshdeep2611:
So I have tried to add this characteristics in the custome component and then expose to apple home but adaptive light feature is not there in my light controller in apple home

PR #441, opened 2023-06-03
Retain the original format of the UUID

bdraco:
apple seems to keep changing the case on these so we will now keep the binary blob and send back the exact bytes that was used in pairing when the client list pairings.

This should fix the users having trouble after migrating to the new home arch

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -25,8 +25,8 @@ import re
 import socket
 import sys
 import tempfile
-import time
 import threading
+import time
 
 from zeroconf import ServiceInfo
 from zeroconf.asyncio import AsyncZeroconf
@@ -41,9 +41,9 @@ from pyhap.const import (
     HAP_REPR_AID,
     HAP_REPR_CHARS,
     HAP_REPR_IID,
-    HAP_REPR_TTL,
     HAP_REPR_PID,
     HAP_REPR_STATUS,
+    HAP_REPR_TTL,
     HAP_REPR_VALUE,
     STANDALONE_AID,
 )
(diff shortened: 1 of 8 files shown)


Ghawken:
Hi 
Thanks for you continued support of these libraries.

With this change here - don't we need hap_handler pair_five also updated?  
Geting an exception as pair_five is sending uuid rather than bytes to accessory_handler.pair?

bdraco:
There is still more. Ran out of time before takeoff.  Will be infrared when finished and tested

bdraco:
I think I got everything. At least mypy checks out.  Traveling though so can't test locally until I get back home

Ghawken:
Thanks.  Happy travels.  Testing here no issues thus far…

bdraco:
all seems ok, but more testing needed and I made some zeroconf changes today so I need to wait to finish testing that for 48 hours before I can validate this to make sure I know which change has which side effects.

speaking of that, this lib needs to accept multiple addresses to publish via mdns since ipv6/ipv4 dual stacks won't work otherwise which would probably make homekit work for a lot more people

Ghawken:
Good luck.
One thing I have recently noticed is pyHAP uses 8.8.8.8 gmail/DNS to decide local ivp4 interface.   If 8.8.8.8 blocked (which many might do to prevent other DNS servers) probably would also lead to issues?

https://github.com/ikalchev/HAP-python/blob/8b62c1965b8d88d79b26d1d857408659ab9b84ba/pyhap/util.py#L48

Interesting still seems to be preferred option - although can use:
`
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        s.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
        s.connect(('10.255.255.255', 80))
        addr = s.getsockname()[0]
`

Which I am testing currently….

ikalchev:
Hey - I will go through the PRs and merge on Sunday probably (not around a computer till then)

bdraco:
Great.  I will try to get the mdns / IPv6 stuff done before than. 

I’m still testing this one

bdraco:
testing looks good

bdraco:
I've been running all 4 of the open PRs on production for a few days.   Everything seems nice and snappy

ikalchev:
@bdraco there seem to be some merge conflicts. Do you mind having a look.

Will merge and release tomorrow if you can address these - that's a promise

bdraco:
I'll fix them in a moment

bdraco:
@ikalchev retested on production after conflicts fixed. all good 👍 

thanks

jimmy927:
Has this made it into main yet ?
Has anyone tested if this fixes https://github.com/home-assistant/core/issues/110540 ?

PR #448, opened 2023-07-23
Improve encryption performance

bdraco:
Every loop would generate a multiple new bytes strings which was quite expensive (more than the encryption itself)

By joining them all together at the end we only have to do it once


bdraco:
@ikalchev Would you be able to get this and https://github.com/ikalchev/HAP-python/pull/447 in a release before Home Assistant beta on Wednesday?

Thank you!

PR #458, opened 2023-10-05
Save raw client username bytes if they are missing on successful pair verify

### LINKED ISSUE #455: Stopped working after upgrading to iOS 17
I have configured a bridge and a single door accessory in home assistant.
They stopped working after updating to iOS 17 my AppleTV & iPhone.

I removed them both from iOS home app and from Home Assistant and tried to reconfigure them as before, but pairing never succeeds. After some minutes, the iPhone shows a message saying that accessory is not reachable and fails.

I am trying to gather the messages from the logs but I recommend avoid upgrading to latest iOS version!

bdraco:
We didn't used to save the raw bytes, but on the success case, iOS is going to send them to us again so we can save them if they are missing

Fixes #455

### DIFF
--- a/pyhap/hap_handler.py
+++ b/pyhap/hap_handler.py
@@ -606,6 +606,13 @@ class HAPServerHandler:
 
         data = tlv.encode(HAP_TLV_TAGS.SEQUENCE_NUM, HAP_TLV_STATES.M4)
         self._send_tlv_pairing_response(data)
+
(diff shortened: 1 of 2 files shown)


bdraco:
@ikalchev Since I was in the area, I fixed I would solve this problem as well as it should allow people who are missing the bytes and have the username saved in the wrong case to avoid re-pairing and it should just start working

PR #454, opened 2023-09-03
Add services and characteristics used by Home Keys

kormax:
This pull request adds following services and characteristics into the `characteristics.json` and `services.json` files:

- AccessoryInformation (Existing):
   * HardwareFinish:   used to retrieve hardware finish (color, material) in order to draw home key pass in a matching color.

- NFCAccess:
   * ConfigurationState
   * NFCAccessSupportedConfiguration: used to retrieve amount of keys that a lock can store.
   * NFCAccessControlPoint: used to configure keys: write/read/delete device/reader keys, etc.

Information about those services had been taken from:
- [@KhaosT - HAP-NodeJS](https://github.com/KhaosT/HAP-NodeJS/commit/80cdb1535f5bee874cc06657ef283ee91f258815);
- [@kupa22 - Apple Home Key](https://github.com/kupa22/apple-homekey)

PR #459, opened 2023-10-05
Add support for Write Responses

kormax:
This PR adds preliminary support for "wr" (Write Response) functionality, where a 'r': True In characteristic update payload indicates a request for 'value' field in response data. You can observe such requests if you run demo code provided by me [in previous PR](https://github.com/ikalchev/HAP-python/pull/454) with DEBUG logging level ON.

This feature is heavily used for Home Key lock configuration, with most commands sent to "[NFCAccessControlPoint (Link to examples)](https://github.com/kupa22/apple-homekey?tab=readme-ov-file#characteristic-nfc-access-control-point)"  expecting to get a special response this way. Without it, the user device will re-try same commands a couple of times, indicating a lack of expected response. (With this commit, behaviour is as expected in a WIP demo app).

Considering my lack of experience with HAP and this library's codebase in particular, I am unable to verify that proposed changes don't break compatibility with existing code (apart from some testing done locally), with an expert's look needed for that. IMO.
My intention was to make as little changes as possible, but IMO `set_characteristics` method could also be refactored to better support this feature in the future. Didn't want to do that due to beforementioned reasons.


ikalchev:
Thanks for the change!

@bdraco can you have a look as well? Looks good to me apart from some needed refactoring at a later stage (not this PR)

bdraco:
Looks ok to me besides this function is too complex and needs to be refactored.  It was already too complex before and this is making it a bit worse.  That can be done in a future PR.

I will test this one production now

bdraco:
Looks ok to me besides this function is too complex and needs to be refactored.  It was already too complex before and this is making it a bit worse.  That can be done in a future PR.

I will test this one production now

bdraco:
Tested on production. No breakage observed (sorry for the multi-posts, I'm on a plane and the wifi is bad)

bdraco:
Tested on production. No breakage observed (sorry for the multi-posts, I'm on a plane and the wifi is bad)

bdraco:
Tested on production. No breakage observed (sorry for the multi-posts, I'm on a plane and the wifi is bad)

kormax:
I think that the only behavior change that could affect legacy code is  that if someone returned a non-None value from setter callback, it would start sending that data as a "value" field in write response. But from what I understand, device should just ignore unknown fields, so that shouldn't be a problem.

PR #462, opened 2023-10-08
Refactor set_characteristics

kormax:
This commit aims at refactoring set_characteristics method, particularly by doing the following
1. Unwrapping multi-layer nested loops into comprehensions and/or single layer loops where possible.
2. Removing as much (if/else) branching as possible
3. Removing (dict) mutations and redundant variables where possible
4. Trying to separate discrete operations into separate loops (for notifications, query setters, accessory + service callbacks)

Some of the changes to the code required modifying 2 tests created by me for write responses, as new behaviour makes it so that no write response value is returned if it wasn't requested, even if callback returned one. This is actually better than before, because this way if someone's legacy code returns a value from a setter by accident, it won't be sent in a response if there was no request for it.

New code seems to pass all other existing unit tests, but lack of regressions definitely has to be verified separately.
If any other changes to the code need to be made (perhaps code style or something else), I am ready to cooperate.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -888,7 +888,8 @@ class AccessoryDriver:
             if HAP_REPR_VALUE not in query and not expired:
                 continue
 
-            aid, iid = query[HAP_REPR_AID], query[HAP_REPR_IID]
+            aid = query[HAP_REPR_AID]
+            iid = query[HAP_REPR_IID]
             value = query.get(HAP_REPR_VALUE)
             write_response_requested = query.get(HAP_REPR_WRITE_RESPONSE, False)
 
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -27,7 +27,7 @@ import sys
 import tempfile
 import threading
 import time
-from typing import Optional, List, Dict, Any, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from zeroconf import ServiceInfo
 from zeroconf.asyncio import AsyncZeroconf
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -879,9 +879,11 @@ class AccessoryDriver:
         updates_by_accessories_services = {}
         char_to_iid = {}
 
+        expired = False
         pid = chars_query.get(HAP_REPR_PID, None)
-        expire_time = self.prepared_writes.get(client_addr, {}).pop(pid, None)
-        expired = False if pid is None else expire_time is None or time.time() > expire_time
+        if pid is not None:
+            expire_time = self.prepared_writes.get(client_addr, {}).pop(pid, None)
+            expired = expire_time is None or time.time() > expire_time
 
         to_update = (query for query in queries if HAP_REPR_VALUE in query or expired)
         for query in to_update:
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -17,6 +17,7 @@ the Characteristic does not block waiting for the actual send to happen.
 """
 import asyncio
 import base64
+from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 import hashlib
 import logging
@@ -865,8 +866,8 @@ class AccessoryDriver:
 
         self._notify(queries, client_addr)
 
-        updates_by_accessories_services = {}
-        results = {}
+        updates_by_accessories_services = defaultdict(lambda: defaultdict(lambda: {}))
+        results = defaultdict(lambda: defaultdict(lambda: {}))
         char_to_iid = {}
 
         expired = False
@@ -905,28 +906,21 @@ class AccessoryDriver:
                 set_result_value is not None and write_response_requested
             )
 
-            if aid not in results:
-                results[aid] = {}
-            if iid not in results[aid]:
-                results[aid][iid] = {}
+            aid_results = results[aid]
 
-            results[aid][iid] = {HAP_REPR_STATUS: set_result}
-
+            aid_results[iid] = {HAP_REPR_STATUS: set_result}
             if return_value_in_response:
-                results[aid][iid][HAP_REPR_VALUE] = set_result_value
+                aid_results[iid][HAP_REPR_VALUE] = set_result_value
 
             char_to_iid[char] = iid
             service = char.service
 
-            if acc not in updates_by_accessories_services:
-                updates_by_accessories_services[acc] = {}
-            if service not in updates_by_accessories_services[acc]:
-                updates_by_accessories_services[acc][service] = {}
             updates_by_accessories_services[acc][service][char] = value
 
         # Proccess accessory and service level setter callbacks
         for acc, updates_by_service in updates_by_accessories_services.items():
             aid = acc.aid
+            aid_results = results[aid]
 
             # Accessory level setter callbacks
             acc_set_result = None
@@ -940,11 +934,11 @@ class AccessoryDriver:
                     char_set_result = _wrap_service_setter(service, chars, client_addr)
                 set_result = char_set_result or acc_set_result
 
-                if set_result:
-                    for char in chars:
-                        results[aid][char_to_iid[char]].update(
-                            {HAP_REPR_STATUS: set_result}
-                        )
+                if not set_result:
+                    continue
+
+                for char in chars:
+                    aid_results[char_to_iid[char]][HAP_REPR_STATUS] = set_result
 
         characteristics = []
         nonempty_results_exist = False
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -27,14 +27,15 @@ import sys
 import tempfile
 import threading
 import time
-from typing import Optional
+from typing import Optional, List, Dict, Any
 
 from zeroconf import ServiceInfo
 from zeroconf.asyncio import AsyncZeroconf
 
 from pyhap import util
 from pyhap.accessory import Accessory, get_topic
-from pyhap.characteristic import CharacteristicError
+from pyhap.characteristic import CharacteristicError, Characteristic
+from pyhap.service import Service
 from pyhap.const import (
     HAP_PERMISSION_NOTIFY,
     HAP_PROTOCOL_SHORT_VERSION,
@@ -862,13 +863,17 @@ class AccessoryDriver:
         """
         # TODO: Add support for chars that do no support notifications.
 
-        queries = chars_query[HAP_REPR_CHARS]
+        queries: List[Dict[str, Any]] = chars_query[HAP_REPR_CHARS]
 
         self._notify(queries, client_addr)
 
-        updates_by_accessories_services = defaultdict(lambda: defaultdict(dict))
-        results = defaultdict(lambda: defaultdict(dict))
-        char_to_iid = {}
+        updates_by_accessories_services: defaultdict[
+            Accessory, defaultdict[Service, dict[Characteristic, Any]]
+        ] = defaultdict(lambda: defaultdict(dict))
+        results: defaultdict[int, defaultdict[int, dict[str, Any]]] = defaultdict(
+            lambda: defaultdict(dict)
+        )
+        char_to_iid: Dict[Characteristic, int] = {}
 
         expired = False
         if HAP_REPR_PID in chars_query:
@@ -884,7 +889,7 @@ class AccessoryDriver:
                 continue
 
             aid, iid = query[HAP_REPR_AID], query[HAP_REPR_IID]
-            value = query.get(HAP_REPR_VALUE, None)
+            value = query.get(HAP_REPR_VALUE)
             write_response_requested = query.get(HAP_REPR_WRITE_RESPONSE, False)
 
             if aid == primary_aid:
@@ -893,28 +898,22 @@ class AccessoryDriver:
                 acc = self.accessory.accessories.get(aid)
             char = acc.get_characteristic(aid, iid)
 
-            set_result, set_result_value = (
-                HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST,
-                None,
-            )
+            set_result = HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST
+            set_result_value = None
+
             if value is not None:
                 set_result, set_result_value = _wrap_char_setter(
                     char, value, client_addr
                 )
 
-            return_value_in_response = (
-                set_result_value is not None and write_response_requested
-            )
+            if set_result_value is not None and write_response_requested:
+                result = {HAP_REPR_STATUS: set_result, HAP_REPR_VALUE: set_result_value}
+            else:
+                result = {HAP_REPR_STATUS: set_result}
 
-            aid_results = results[aid]
-
-            aid_results[iid] = {HAP_REPR_STATUS: set_result}
-            if return_value_in_response:
-                aid_results[iid][HAP_REPR_VALUE] = set_result_value
-
+            results[aid][iid] = result
             char_to_iid[char] = iid
             service = char.service
-
             updates_by_accessories_services[acc][service][char] = value
 
         # Proccess accessory and service level setter callbacks
@@ -1001,11 +1000,10 @@ class AccessoryDriver:
         for query in queries:
             if HAP_PERMISSION_NOTIFY not in query:
                 continue
-            aid, iid, ev = (
-                query[HAP_REPR_AID],
-                query[HAP_REPR_IID],
-                query[HAP_PERMISSION_NOTIFY],
-            )
+            aid = query[HAP_REPR_AID]
+            iid = query[HAP_REPR_IID]
+            ev = query[HAP_PERMISSION_NOTIFY]
+
             char_topic = get_topic(aid, iid)
             action = "Subscribed" if ev else "Unsubscribed"
             logger.debug("%s client %s to topic %s", action, client_addr, char_topic)
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -35,7 +35,6 @@ from zeroconf.asyncio import AsyncZeroconf
 from pyhap import util
 from pyhap.accessory import Accessory, get_topic
 from pyhap.characteristic import Characteristic, CharacteristicError
-from pyhap.service import Service
 from pyhap.const import (
     HAP_PERMISSION_NOTIFY,
     HAP_PROTOCOL_SHORT_VERSION,
@@ -55,6 +54,7 @@ from pyhap.hap_server import HAPServer
 from pyhap.hsrp import Server as SrpServer
 from pyhap.loader import Loader
 from pyhap.params import get_srp_context
+from pyhap.service import Service
 from pyhap.state import State
 
 from .const import HAP_SERVER_STATUS
@@ -997,8 +997,10 @@ class AccessoryDriver:
             logger.error("Could not stop AccessoryDriver because of error: %s", e)
             raise
 
-    def _notify(self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]) -> None:
-        """Notify the driver that the client has subscribed or unsubscribed."""    
+    def _notify(
+        self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]
+    ) -> None:
+        """Notify the driver that the client has subscribed or unsubscribed."""
         for query in queries:
             if HAP_PERMISSION_NOTIFY not in query:
                 continue
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -922,7 +922,7 @@ class AccessoryDriver:
                 updates_by_accessories_services[acc] = {}
             if service not in updates_by_accessories_services[acc]:
                 updates_by_accessories_services[acc][service] = {}
-            updates_by_accessories_services[acc][service].update({char: value})
+            updates_by_accessories_services[acc][service][char] = value
 
         # Proccess accessory and service level setter callbacks
         for acc, updates_by_service in updates_by_accessories_services.items():
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -27,7 +27,7 @@ import sys
 import tempfile
 import threading
 import time
-from typing import Optional, List, Dict, Any
+from typing import Optional, List, Dict, Any, Tuple
 
 from zeroconf import ServiceInfo
 from zeroconf.asyncio import AsyncZeroconf
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -997,7 +997,7 @@ class AccessoryDriver:
             logger.error("Could not stop AccessoryDriver because of error: %s", e)
             raise
 
-    def _notify(self, queries, client_addr):
+    def _notify(self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]) -> None:
         for query in queries:
             if HAP_PERMISSION_NOTIFY not in query:
                 continue
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -998,6 +998,7 @@ class AccessoryDriver:
             raise
 
     def _notify(self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]) -> None:
+        """Notify the driver that the client has subscribed or unsubscribed."""    
         for query in queries:
             if HAP_PERMISSION_NOTIFY not in query:
                 continue
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -920,7 +920,7 @@ class AccessoryDriver:
                 set_result = _wrap_acc_setter(acc, updates_by_service, client_addr)
 
                 characteristics = (
-                    char for chars in updates_by_accessories_services[acc].values()
+                    char for chars in updates_by_service.values()
                     for char in chars.keys()
                 )
                 for char in characteristics:
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -866,13 +866,13 @@ class AccessoryDriver:
 
         self._notify(queries, client_addr)
 
-        updates_by_accessories_services = defaultdict(lambda: defaultdict(lambda: {}))
-        results = defaultdict(lambda: defaultdict(lambda: {}))
+        updates_by_accessories_services = defaultdict(lambda: defaultdict(dict))
+        results = defaultdict(lambda: defaultdict(dict))
         char_to_iid = {}
 
         expired = False
-        pid = chars_query.get(HAP_REPR_PID, None)
-        if pid is not None:
+        if HAP_REPR_PID in chars_query:
+            pid = chars_query[HAP_REPR_PID]
             expire_time = self.prepared_writes.get(client_addr, {}).pop(pid, None)
             expired = expire_time is None or time.time() > expire_time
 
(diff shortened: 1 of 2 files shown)


kormax:
Reported coverage decreased because method size decreased in comparison to the whole file.
`set_characteristics` as indicated by report is fully covered

bdraco:
Looks like there is an extra dict lookup
pyhap/accessory_driver.py:923:38: R1733: Unnecessary dictionary index lookup, use 'updates_by_service' instead (unnecessary-dict-index-lookup)

kormax:
Thanks for finding that out, a peculiar mistake. 
Strangely, my local pylint didn't report that. Perhaps I called it with wrong parameters?

kormax:
Having taken a look at code once again, I think there's also a possibility of doing accessory and service level callbacks in one loop (remove multiple loops over chars), but i'll only be able to make those changes in a couple of hours.

bdraco:
Take your time. Its super later here for me (GMT-10) so I won't have a chance to look at it again until tomorrow anyways

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,99 @@ def set_characteristics(self, chars_query, client_addr):
...
 
-            updates_by_service = {}
-            char_to_iid = {}
-            for iid, value in new_iid_values.items():
-                # Characteristic level setter callbacks
-                char = acc.get_characteristic(aid, iid)
+            updates_by_accessories_services \
+                .setdefault(acc, {}) \
+                .setdefault(char.service, {}) \
+                .update({char: value})
```
bdraco, line 915: Please format with `black`

Also `setdefault` still creates an empty dict and throws it away.  We should avoid doing that in a loop

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,99 @@ def set_characteristics(self, chars_query, client_addr):
...
+        to_update = (query for query in queries if HAP_REPR_VALUE in query or expired)
+        for query in to_update:
+            aid, iid = query[HAP_REPR_AID], query[HAP_REPR_IID]
+            value = query.get(HAP_REPR_VALUE, None)
+            write_response_requested = query.get(HAP_REPR_WRITE_RESPONSE, False)
 
-            if HAP_REPR_WRITE_RESPONSE in cq:
-                setter_responses.setdefault(aid, {})
-                had_write_response = True
+            acc = self.accessory if self.accessory.aid == aid else None
```
bdraco, line 894: This code is performance sensitive. Please unpack the accessory aid into a local var outside the loop so we don't have to look it up each loop

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,99 @@ def set_characteristics(self, chars_query, client_addr):
...
+                **callback_results.get(aid, {}).get(iid, {})
+            }
+            for aid, query_iid_results in query_results.items()
+            for iid, query_results in query_iid_results.items()
+        ]
+        nonempty_results_exist = any(
+            result[HAP_REPR_STATUS] != HAP_SERVER_STATUS.SUCCESS
+            or HAP_REPR_VALUE in result
+            for result in results
+        )
```
bdraco, line 954: Its probably a bit more readable (and faster) to unpack this into a single loop since the generator expression is having to iterate everything again

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,99 @@ def set_characteristics(self, chars_query, client_addr):
...
+        for query in to_notify:
+            aid, iid, ev = query[HAP_REPR_AID], query[HAP_REPR_IID], query[HAP_PERMISSION_NOTIFY]
+            char_topic = get_topic(aid, iid)
+            action = "Subscribed" if ev else "Unsubscribed"
+            logger.debug(
+                "%s client %s to topic %s", action, client_addr, char_topic
+            )
+            self.async_subscribe_client_topic(
+                client_addr, char_topic, ev
+            )
```
bdraco, line 876: This looks fully self contained. I think it could be broken out into another function to make this function smaller and easier to read

pyhap/accessory_driver.py
```diff
@@ -67,12 +67,13 @@
 VALID_MDNS_REGEX = re.compile(r"[^A-Za-z0-9\-]+")
 LEADING_TRAILING_SPACE_DASH = re.compile(r"^[ -]+|[ -]+$")
 DASH_REGEX = re.compile(r"[-]+")
+KEYS_TO_EXCLUDE = set((HAP_REPR_IID, HAP_REPR_AID))
```
bdraco, line 70:
```suggestion
KEYS_TO_EXCLUDE = {HAP_REPR_IID, HAP_REPR_AID}
```

bdraco: Probably would have been caught by `pyupgrade` anyways

bdraco:
Thanks @kormax 

Some small feedback above.

Cheers

pyhap/accessory_driver.py
kormax: Cannot be done here, as we process queries in a flat loop, so there's no guarantee that next iteration has the same accessory. (As far as I understand).

pyhap/accessory_driver.py
kormax: Makes sense. Protected class method would be fine for that?

pyhap/accessory_driver.py
kormax: BTW. If you're looking at changes at GitHub, it has massacred the code view. I'd advice looking at the differences side by side.

bdraco: untested, but I was thinking something like this

```diff
diff --git a/pyhap/accessory_driver.py b/pyhap/accessory_driver.py
index e6fa39c..6265ec6 100644
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -885,14 +885,18 @@ class AccessoryDriver:
             expire_time = self.prepared_writes.get(client_addr, {}).pop(pid, None)
             expired = expire_time is None or time.time() > expire_time
 
-        to_update = (query for query in queries if HAP_REPR_VALUE in query or expired)
-        for query in to_update:
+        primary_accessory = self.accessory
+        primary_aid = primary_accessory.aid
+
+        for query in queries:
+            if expired or HAP_REPR_VALUE not in query:
+                continue
             aid, iid = query[HAP_REPR_AID], query[HAP_REPR_IID]
             value = query.get(HAP_REPR_VALUE, None)
             write_response_requested = query.get(HAP_REPR_WRITE_RESPONSE, False)
 
-            acc = self.accessory if self.accessory.aid == aid else None
-            acc = acc or self.accessory.accessories.get(aid)
+            acc = primary_accessory if primary_aid == aid else None
+            acc = acc or primary_accessory.accessories.get(aid)
 
             char = acc.get_characteristic(aid, iid)
 
```

bdraco: need to double check the `expire` handling here

kormax: Oh, I didn't understand you initially then. This makes sense, I need to test.

pyhap/accessory_driver.py
kormax: Gonna look into that.

pyhap/accessory_driver.py
kormax: I have a question regarding the `continue` vs `generator`. Is this more of stylistic choice of this project, or are there any performance considerations?

bdraco: It's a performance concern.  This particular function is very hot when users are using it with lighting in Home Assistant

pyhap/accessory_driver.py
kormax: Ran a benchmark via pytest-benchmark. On low input sizes (<10) list-append-based solution indeed has a ~1us mean gain (M1 Pro) over this one. 

Didn't consider that queries are not that big usually, so append solution does not cause a resize. Gonna change.

pyhap/accessory_driver.py
kormax: Sorry for that. If I knew, I would've approached this in other way initially (apart from genexps, which I assumed ran faster that continue). 

I'm unsure if this PR as a whole does not form a regression. I think it would be wise to test it against original method.

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,106 @@ def set_characteristics(self, chars_query, client_addr):
...
+                results[aid][iid][HAP_REPR_VALUE] = set_result_value
+
+            char_to_iid[char] = iid
+            service = char.service
+
+            if acc not in updates_by_accessories_services:
+                updates_by_accessories_services[acc] = {}
+            if service not in updates_by_accessories_services[acc]:
+                updates_by_accessories_services[acc][service] = {}
+            updates_by_accessories_services[acc][service].update({char: value})
```
bdraco, line 925: You can set item aka [] here instead of update

kormax:
Hello.

I've pushed a new commit (at the moment of writing it is being tested), with changes proposed by you, plus a couple of changes added by me considering the importance of performance:
1. `setdefault`, `get` methods were replaced with `in` checks for initial value creation
2. `continue` returned in place of `genexp`
3. replace some dict destructuring with mutations
4. unify query_results and callback_results, to reduce memory taken and time used on 
5. move notify logic to separate method

I think I've forgotten something, will change if find out what.

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,106 @@ def set_characteristics(self, chars_query, client_addr):
...
-                    continue
-                char_to_iid[char] = iid
-                updates_by_service.setdefault(char.service, {}).update({char: value})
+            if return_value_in_response:
+                results[aid][iid][HAP_REPR_VALUE] = set_result_value
+
+            char_to_iid[char] = iid
+            service = char.service
+
+            if acc not in updates_by_accessories_services:
```
bdraco, line 921: I think you could avoid this by using collections defaultdict since it implemented in native code and fast

pyhap/accessory_driver.py
```diff
@@ -859,118 +860,106 @@ def set_characteristics(self, chars_query, client_addr):
...
 
-                if set_result_value is not None:
-                    if setter_responses.get(aid, None) is None:
-                        logger.warning(
-                            "Returning write response '%s' when it wasn't requested for %s %s",
-                            set_result_value, aid, iid
-                        )
-                    had_write_response = True
-                    setter_responses.setdefault(aid, {})[iid] = set_result_value
+            results[aid][iid] = {HAP_REPR_STATUS: set_result}
```
bdraco, line 913: I would assign the new dict to a local var and than set results[aid][iid] = var at the end of the loop

This avoids all the lookups to find the inner dict each time you set a key/value pair

pyhap/accessory_driver.py
kormax: Good idea. Thanks

pyhap/accessory_driver.py
```diff
@@ -859,118 +861,99 @@ def set_characteristics(self, chars_query, client_addr):
...
+        updates_by_accessories_services = defaultdict(lambda: defaultdict(lambda: {}))
+        results = defaultdict(lambda: defaultdict(lambda: {}))
+        char_to_iid = {}
 
-            if expired:
-                setter_results[aid][iid] = HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST
-                had_error = True
-                continue
+        expired = False
+        pid = chars_query.get(HAP_REPR_PID, None)
```
bdraco, line 874:
```suggestion
        pid = chars_query.get(HAP_REPR_PID)
```

`None` is already the default

pyhap/accessory_driver.py
```diff
@@ -859,118 +861,99 @@ def set_characteristics(self, chars_query, client_addr):
...
-                if not char.service or (
-                    not acc.setter_callback and not char.service.setter_callback
-                ):
-                    continue
-                char_to_iid[char] = iid
-                updates_by_service.setdefault(char.service, {}).update({char: value})
+            set_result, set_result_value = (
+                HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST,
+                None,
+            )
```
bdraco, line 899:
```suggestion
            set_result = HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST
            set_result_value = None
```

kormax: This was done by black

pyhap/accessory_driver.py
```diff
@@ -859,118 +861,99 @@ def set_characteristics(self, chars_query, client_addr):
...
+            return_value_in_response = (
+                set_result_value is not None and write_response_requested
+            )
+
+            aid_results = results[aid]
+
+            aid_results[iid] = {HAP_REPR_STATUS: set_result}
+            if return_value_in_response:
+                aid_results[iid][HAP_REPR_VALUE] = set_result_value
+
```
bdraco:
```suggestion
            aid_iid_results = {HAP_REPR_STATUS: set_result}
            if return_value_in_response:
                aid_iid_results[HAP_REPR_VALUE] = set_result_value
            results[aid][iid] = aid_iid_resultss[iid][HAP_REPR_VALUE] = set_result_value

```

pyhap/accessory_driver.py
kormax: I reverted expired calculation to original format

pyhap/accessory_driver.py
bdraco: black may complain about this suggestion

kormax: Maybe even replace it with branched initialisation? One with value and one without?

pyhap/accessory_driver.py
bdraco: Maybe there was a comma out of place? I can't replicate that

```
bdraco@MacBook-Pro-9 HAP-python % black pyhap/accessory_driver.py
All done! ✨ 🍰 ✨
1 file left unchanged.
bdraco@MacBook-Pro-9 HAP-python % git diff -w
diff --git a/pyhap/accessory_driver.py b/pyhap/accessory_driver.py
index 7bb0984..6cc6966 100644
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -893,10 +893,9 @@ class AccessoryDriver:
                 acc = self.accessory.accessories.get(aid)
             char = acc.get_characteristic(aid, iid)
 
-            set_result, set_result_value = (
-                HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST,
-                None,
-            )
+            set_result = HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST
+            set_result_value = None
+
             if value is not None:
                 set_result, set_result_value = _wrap_char_setter(
                     char, value, client_addr
@@ -906,12 +905,10 @@ class AccessoryDriver:
                 set_result_value is not None and write_response_requested
             )
 
-            aid_results = results[aid]
-
-            aid_results[iid] = {HAP_REPR_STATUS: set_result}
+            aid_iid_results = {HAP_REPR_STATUS: set_result}
             if return_value_in_response:
-                aid_results[iid][HAP_REPR_VALUE] = set_result_value
-
+                aid_iid_results[HAP_REPR_VALUE] = set_result_value
+            results[aid][iid] = aid_iid_results
             char_to_iid[char] = iid
             service = char.service
 
```

kormax: Sorry, I've misread this code as having both inits in the same line. Doing it on separate - ok.

bdraco:
The nesting is really hard to follow.

Maybe some typing will make it easier to review in VSCode
```diff
diff --git a/pyhap/accessory_driver.py b/pyhap/accessory_driver.py
index 7bb0984..34ba3b1 100644
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -17,6 +17,7 @@ the Characteristic does not block waiting for the actual send to happen.
 """
 import asyncio
 import base64
+from typing import Any
 from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 import hashlib
@@ -27,14 +28,15 @@ import sys
 import tempfile
 import threading
 import time
-from typing import Optional
+from typing import Optional, List, Dict
 
 from zeroconf import ServiceInfo
 from zeroconf.asyncio import AsyncZeroconf
 
 from pyhap import util
 from pyhap.accessory import Accessory, get_topic
-from pyhap.characteristic import CharacteristicError
+from pyhap.characteristic import CharacteristicError, Characteristic
+from pyhap.service import Service
 from pyhap.const import (
     HAP_PERMISSION_NOTIFY,
     HAP_PROTOCOL_SHORT_VERSION,
@@ -862,13 +864,17 @@ class AccessoryDriver:
         """
         # TODO: Add support for chars that do no support notifications.
 
-        queries = chars_query[HAP_REPR_CHARS]
+        queries: List[Dict[str, Any]] = chars_query[HAP_REPR_CHARS]
 
         self._notify(queries, client_addr)
 
-        updates_by_accessories_services = defaultdict(lambda: defaultdict(lambda: {}))
-        results = defaultdict(lambda: defaultdict(lambda: {}))
-        char_to_iid = {}
+        updates_by_accessories_services: defaultdict[
+            Accessory, defaultdict[Service, dict[Characteristic, Any]]
+        ] = defaultdict(lambda: defaultdict(lambda: {}))
+        results: defaultdict[int, defaultdict[int, dict[str, Any]]] = defaultdict(
+            lambda: defaultdict(lambda: {})
+        )
+        char_to_iid: Dict[Characteristic, int] = {}
 
         expired = False
         pid = chars_query.get(HAP_REPR_PID, None)
@@ -893,10 +899,9 @@ class AccessoryDriver:
                 acc = self.accessory.accessories.get(aid)
             char = acc.get_characteristic(aid, iid)
 
-            set_result, set_result_value = (
-                HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST,
-                None,
-            )
+            set_result = HAP_SERVER_STATUS.INVALID_VALUE_IN_REQUEST
+            set_result_value = None
+
             if value is not None:
                 set_result, set_result_value = _wrap_char_setter(
                     char, value, client_addr
@@ -906,12 +911,10 @@ class AccessoryDriver:
                 set_result_value is not None and write_response_requested
             )
 
-            aid_results = results[aid]
-
-            aid_results[iid] = {HAP_REPR_STATUS: set_result}
+            aid_iid_results = {HAP_REPR_STATUS: set_result}
             if return_value_in_response:
-                aid_results[iid][HAP_REPR_VALUE] = set_result_value
-
+                aid_iid_results[HAP_REPR_VALUE] = set_result_value
+            results[aid][iid] = aid_iid_results
             char_to_iid[char] = iid
             service = char.service
 
@@ -940,7 +943,7 @@ class AccessoryDriver:
                 for char in chars:
                     aid_results[char_to_iid[char]][HAP_REPR_STATUS] = set_result
 
-        characteristics = []
+        characteristics: List[Dict[str, Any]] = []
         nonempty_results_exist = False
         for aid, iid_results in results.items():
             for iid, result in iid_results.items():
```

kormax:
You do type hints in this project? I saw none so didn't add them.

bdraco:
> You do type hints in this project? I saw none so didn't add them assuming that was intended. I can of course do this.

I've been slowly adding them. Emphasis on slowly 😄

pyhap/accessory_driver.py
```diff
@@ -859,118 +861,99 @@ def set_characteristics(self, chars_query, client_addr):
...
-            if HAP_PERMISSION_NOTIFY in cq:
-                char_topic = get_topic(aid, iid)
-                action = "Subscribed" if cq[HAP_PERMISSION_NOTIFY] else "Unsubscribed"
-                logger.debug(
-                    "%s client %s to topic %s", action, client_addr, char_topic
-                )
-                self.async_subscribe_client_topic(
-                    client_addr, char_topic, cq[HAP_PERMISSION_NOTIFY]
+            aid, iid = query[HAP_REPR_AID], query[HAP_REPR_IID]
+            value = query.get(HAP_REPR_VALUE, None)
```
bdraco, line 887:
```suggestion
            value = query.get(HAP_REPR_VALUE)
```

pyhap/accessory_driver.py
```diff
@@ -1013,3 +996,17 @@ def signal_handler(self, _signal, _frame):
...
+
+    def _notify(self, queries, client_addr):
+        for query in queries:
+            if HAP_PERMISSION_NOTIFY not in query:
+                continue
+            aid, iid, ev = (
+                query[HAP_REPR_AID],
+                query[HAP_REPR_IID],
+                query[HAP_PERMISSION_NOTIFY],
+            )
```
bdraco, line 1008:
```suggestion
            aid = query[HAP_REPR_AID]
            iid = query[HAP_REPR_IID]
            ev = query[HAP_PERMISSION_NOTIFY]
```

pyhap/accessory_driver.py
```diff
@@ -859,118 +861,99 @@ def set_characteristics(self, chars_query, client_addr):
...
 
-            if HAP_PERMISSION_NOTIFY in cq:
-                char_topic = get_topic(aid, iid)
-                action = "Subscribed" if cq[HAP_PERMISSION_NOTIFY] else "Unsubscribed"
-                logger.debug(
-                    "%s client %s to topic %s", action, client_addr, char_topic
-                )
-                self.async_subscribe_client_topic(
-                    client_addr, char_topic, cq[HAP_PERMISSION_NOTIFY]
+            aid, iid = query[HAP_REPR_AID], query[HAP_REPR_IID]
```
bdraco, line 886:
```suggestion
            aid = query[HAP_REPR_AID]
            iid = query[HAP_REPR_IID]
```

pyhap/accessory_driver.py
kormax: This one breaks code.

kormax: Sorry, my mistake. Didn't know get doesn't throw if no default is given.

bdraco: No worries. That's why we have code review 😄

kormax:
We have a problem with typehints, 3.8 didn't allow to use most builtin types with subscripts as hints. Perhaps replace with simple dict?

pyhap/accessory_driver.py
```diff
@@ -859,118 +862,98 @@ def set_characteristics(self, chars_query, client_addr):
...
+        queries: List[Dict[str, Any]] = chars_query[HAP_REPR_CHARS]
+
+        self._notify(queries, client_addr)
+
+        updates_by_accessories_services: defaultdict[
+            Accessory, defaultdict[Service, dict[Characteristic, Any]]
+        ] = defaultdict(lambda: defaultdict(dict))
+        results: defaultdict[int, defaultdict[int, dict[str, Any]]] = defaultdict(
+            lambda: defaultdict(dict)
+        )
```
bdraco, line 875:
```suggestion
        updates_by_accessories_services: Dict[
            Accessory, Dict[Service, Dict[Characteristic, Any]]
        ] = defaultdict(lambda: defaultdict(dict))
        results: Dict[int, Dict[int, dict[str, Any]]] = defaultdict(
            lambda: defaultdict(dict)
        )
```

`defaultdict` wasn't subscriptable until py3.9 sadly so we have to use `Dict` since we still support legacy python

pyhap/accessory_driver.py
```diff
@@ -859,118 +862,98 @@ def set_characteristics(self, chars_query, client_addr):
...
-        expired = False
 
+        queries: List[Dict[str, Any]] = chars_query[HAP_REPR_CHARS]
+
+        self._notify(queries, client_addr)
+
+        updates_by_accessories_services: Dict[
+            Accessory, Dict[Service, Dict[Characteristic, Any]]
+        ] = defaultdict(lambda: defaultdict(dict))
+        results: Dict[int, Dict[int, dict[str, Any]]] = defaultdict(
```
bdraco, line 873:
```suggestion
        results: Dict[int, Dict[int, Dict[str, Any]]] = defaultdict(
```

bdraco: one more (I always forget these since I'm used to working with py3.11 which has so much better typing!)

pyhap/accessory_driver.py
```diff
@@ -1013,3 +996,16 @@ def signal_handler(self, _signal, _frame):
         except Exception as e:
             logger.error("Could not stop AccessoryDriver because of error: %s", e)
             raise
+
+    def _notify(self, queries, client_addr):
```
bdraco, line 1000:
```suggestion
    def _notify(self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]) -> None:
```

Will need `Tuple` import above

kormax: Could even be `Iterable` as far as I can see.

pyhap/accessory_driver.py
```diff
@@ -1013,3 +996,16 @@ def signal_handler(self, _signal, _frame):
         except Exception as e:
             logger.error("Could not stop AccessoryDriver because of error: %s", e)
             raise
+
+    def _notify(self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]) -> None:
```
bdraco, line 1000:
```suggestion
    def _notify(self, queries: List[Dict[str, Any]], client_addr: Tuple[str, int]) -> None:
        """Notify the driver that the client has subscribed or unsubscribed."""    
```

pyhap/accessory_driver.py
```diff
@@ -26,14 +27,15 @@
 import tempfile
 import threading
 import time
-from typing import Optional
+from typing import Optional, List, Dict, Any
```
bdraco, line 30:
```suggestion
from typing import Optional, List, Dict, Any, Tuple
```

tests/test_accessory_driver.py
```diff
@@ -177,16 +177,7 @@ def setter_with_write_response(value=0):
...
-        HAP_REPR_CHARS: [
-            {
-                HAP_REPR_AID: acc.aid,
-                HAP_REPR_IID: char_nfc_access_control_point_iid,
-                HAP_REPR_STATUS: 0,
-                HAP_REPR_VALUE: 1
-            },
-        ]
-    }
+    assert response is None
```
bdraco, line 180: Was this a bug?

kormax: No, this was a change of behaviour made by me, I described it in a first post.

Original PR that added write responses could cause side effects for third-party code whose setters returned values on accident (when it wasn't a feature). In this PR I added a check that response value is only sent back if it was requested.

pyhap/accessory_driver.py
```diff
@@ -26,14 +27,15 @@
 import tempfile
 import threading
 import time
-from typing import Optional
+from typing import Optional, List, Dict, Any, Tuple
```
bdraco, line 30:
```suggestion
from typing import Any, Dict, List, Optional, Tuple
```

pyhap/accessory_driver.py
```diff
@@ -26,14 +27,15 @@
...
 
 from zeroconf import ServiceInfo
 from zeroconf.asyncio import AsyncZeroconf
 
 from pyhap import util
 from pyhap.accessory import Accessory, get_topic
-from pyhap.characteristic import CharacteristicError
+from pyhap.characteristic import CharacteristicError, Characteristic
```
bdraco, line 37:
```suggestion
from pyhap.characteristic import Characteristic, CharacteristicError
```

pyhap/accessory_driver.py
```diff
@@ -26,14 +27,15 @@
...
+from pyhap.service import Service
```
bdraco, line 38: Needs go below `pyhap.params` import

Or you can do

`isort pyhap/accessory_driver.py`
`black pyhap/accessory_driver.py`

bdraco:
Looks good.

I think we just have to make the linter happy, and than I'll push this to my production to start testing

bdraco:
Great.  I'll push this to my production setup and report back

bdraco:
Everything appears to be working as normal

No traces in the log.

Will let it run for 48 hours and report back again

kormax:
Many thanks for your work.

bdraco:
Thank you!

Performance testing also looks good 👍

kormax:
May I ask you how you're doing it. Didn't find any references to any tools in the project, perhaps I missed it. 
Will use it in the future before making some proposals.

bdraco:
> May I ask you how you're doing it?
> 
> Didn't find any references to any perf analysis tools in the project, perhaps I missed it. Will use it in the future before making some proposals.

Updated Home Assistant with the new changes

Ran https://www.home-assistant.io/integrations/profiler/#service-profilerstart for 60s while flipping a light on and off over and over

Compared profile before and after

Did the same this with a `py-spy`

bdraco (approved):
LGTM, testing 👍

PR #464, opened 2023-10-11
Fix hashing of accessories to not include the value

bdraco:
~~Needs https://github.com/ikalchev/HAP-python/pull/463 merged first~~

The hash of accessories would include the value which meant that we would increment the `c#` very frequently which could drain battery and force iCloud to sync over and over.  The hash should only include the actual config and not the value.  The net result is everything could appear unavailable since the iCloud sync would never finish for some users and the constant `get_accessories` could overload the system. 

HA just added support for [reloading accessories](https://github.com/home-assistant/core/pull/101719) which will generate a lot more get requests since the `c#` value is going to change more frequently.

`to_HAP` is now cached to handle this.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,3 +4,19 @@ include = pyhap/*
 omit =
   tests/*
   pyhap/accessories/*
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain about missing debug-only code:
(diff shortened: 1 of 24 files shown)


bdraco:
Leaving as draft until parent merges

bdraco:
We have a bug here, we need the hash without the value for the `c#`

bdraco:
Fixing conflicts...

bdraco:
Retesting

bdraco:
All good

PR #467, opened 2023-10-22
Fix handling of explict close

bdraco:
The check for next_event was not correct for explict closes as https://github.com/python-hyper/h11/blob/a2c68948accadc3876dffcf979d98002e4a4ed27/h11/_connection.py#L445

will only return h11.ConnectionClosed as an object and not a type

### DIFF
--- a/pyhap/hap_protocol.py
+++ b/pyhap/hap_protocol.py
@@ -46,7 +46,7 @@ class HAPServerProtocol(asyncio.Protocol):
         connections: Dict[str, "HAPServerProtocol"],
         accessory_driver: "AccessoryDriver",
     ) -> None:
-        self.loop: asyncio.AbstractEventLoop = loop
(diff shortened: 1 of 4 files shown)


pyhap/hap_protocol.py
```diff
@@ -246,27 +246,33 @@ def _process_one_event(self) -> bool:
         logger.debug(
             "%s (%s): h11 Event: %s", self.peername, self.handler.client_uuid, event
         )
-        if event in (h11.NEED_DATA, h11.ConnectionClosed):
```
bdraco, line 249: This bug was missed because there was no explicit test and it looked like it was covered because we did match on NEED_DATA

bdraco:
Thanks. It would probably be good to get a patch release for this since it can leave around connections otherwise

PR #470, opened 2023-11-27
Late-import base36 and QR code libraries; remove SUPPORT_QR_CODE flag

akx:
Some tests run locally for `home-assistant/core` would fail with a cryptic "base36 is not defined" error if that library was not installed.

This PR:

* makes that error message cleaner – it's now a `RuntimeError` that has the import error as a reason
* makes `pyhap` not import `base36` or `QRCode` at all until they are needed 
  * in doing that removes the `SUPPORT_QR_CODE` flag; based on [GitHub Code Search](https://github.com/search?q=SUPPORT_QR_CODE+language%3APython&type=code&l=Python) there are no external users for that flag that I can see.


akx:
Ping @ikalchev :)

ikalchev:
Sorry for the huge delay on that!

PR #471, opened 2024-04-21
Avoid os.chmod failing on Windows if file non-existant

pjkundert:
On Windows, os.chmod fails if the target file doesn't exist.  Avoid calling it in that case.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -642,16 +642,18 @@ class AccessoryDriver:
         tmp_filename = None
         try:
             temp_dir = os.path.dirname(self.persist_file)
+            logger.debug(f"Creating temp persist file in {temp_dir!r}")
             with tempfile.NamedTemporaryFile(
                 mode="w", dir=temp_dir, delete=False
             ) as file_handle:
                 tmp_filename = file_handle.name
+                logger.debug(f"Created temp persist file {file_handle!r} named {tmp_filename!r}")
                 self.encoder.persist(file_handle, self.state)
             if (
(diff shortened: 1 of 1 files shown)


pyhap/accessory_driver.py
```diff
@@ -642,16 +642,18 @@ def persist(self):
         tmp_filename = None
         try:
             temp_dir = os.path.dirname(self.persist_file)
+            logger.debug(f"Creating temp persist file in {temp_dir!r}")
```
bdraco, line 645: Please use lazy formatting for loggers
https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html

pyhap/accessory_driver.py
```diff
@@ -642,16 +642,18 @@ def persist(self):
...
             with tempfile.NamedTemporaryFile(
                 mode="w", dir=temp_dir, delete=False
             ) as file_handle:
                 tmp_filename = file_handle.name
+                logger.debug(f"Created temp persist file {file_handle!r} named {tmp_filename!r}")
```
bdraco, line 650: Please use lazy formatting for loggers
https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html

pyhap/accessory_driver.py
ikalchev: fixed it

pyhap/accessory_driver.py
ikalchev: fixed it

PR #476, opened 2024-11-01
Implement zerocopy writes for the encrypted protocol

bdraco:
With Python 3.12+ and later `transport.writelines` is implemented as [`sendmsg(..., IOV_MAX)`](https://github.com/python/cpython/issues/91166) which allows us to avoid joining the bytes and sending them in one go.

Older Python will effectively do the same thing we do now `b"".join(...)`

### DIFF
--- a/pyhap/hap_crypto.py
+++ b/pyhap/hap_crypto.py
@@ -3,7 +3,7 @@ from functools import partial
 import logging
 import struct
 from struct import Struct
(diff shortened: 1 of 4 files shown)


bdraco:
@ikalchev When you get a chance, I have a few PRs in the queue

ikalchev:
Awesome stuff! 🙏

PR #481, opened 2024-11-03
Address remaining pylint complaints

bdraco:
Note that CI will still fail because codecov now requires an upload token.  We end up hard coding them in ` .codecov.yml` for most repos now ex https://github.com/aio-libs/yarl/blob/dfc41dbfcd5a91283ab6c5c21d70bc4b1c377789/.codecov.yml#L10

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -323,7 +323,7 @@ class AccessoryDriver:
                 and os.name != "nt"
             ):
                 logger.debug("Setting child watcher")
-                watcher = asyncio.SafeChildWatcher()
+                watcher = asyncio.SafeChildWatcher() # pylint: disable=deprecated-class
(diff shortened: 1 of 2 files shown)


pyhap/accessory_driver.py
```diff
@@ -323,7 +323,7 @@ def start(self):
                 and os.name != "nt"
             ):
                 logger.debug("Setting child watcher")
-                watcher = asyncio.SafeChildWatcher()
+                watcher = asyncio.SafeChildWatcher() # pylint: disable=deprecated-class
```
bdraco, line 326: Probably delete this once support for Python 3.11 and older is dropped

ikalchev:
Yeah I know I know. I have little time these days, unfortunately, but will try to address such maintenance issues son. Thanks for doing a stellar job!

PR #486, opened 2025-06-16
Replace deprecated asyncio.iscoroutinefunction

cdce8p:
Python 3.14 will deprecated `asyncio.iscoroutinefunction` in favor of `inspect.iscoroutinefunction` which has been available for some time now.

https://docs.python.org/3.14/deprecations/pending-removal-in-3.16.html
https://docs.python.org/3.14/library/inspect.html#inspect.iscoroutinefunction

### DIFF
--- a/pyhap/util.py
+++ b/pyhap/util.py
@@ -1,6 +1,7 @@
 import asyncio
 import base64
 import functools
+import inspect
 import random
 import socket
 import sys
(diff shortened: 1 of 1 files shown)


cdce8p:
@ikalchev Just rebased this one, it's ready now. Would be great if it could be included in the next release as well.

PR #493, opened 2026-07-04
Add classic HomeKit Secure Video recording

RaHehl:
## Summary

Adds **HomeKit Secure Video (classic HKSV)** to the `Camera` accessory: an accessory can now record event-triggered fragmented-MP4 clips to a HomeKit Home Hub over the HomeKit Data Stream (HDS).

Verified end-to-end against a real iOS 26.5 Home Hub — motion triggers the hub to open a recording stream and pull fMP4 fragments over the encrypted HDS connection.

Non-breaking and opt-in: existing cameras are unaffected; recording is enabled via a `recording` option.

## What's included

- **HDS transport** — framing + ChaCha20-Poly1305 key derivation (`hds.py`), the opack message codec (`hds_protocol.py`), the TCP listener/dispatch (`hds_server.py`), the `dataSend` recording fragment transfer (`hds_recording.py`).
- **Recording configuration TLV codecs** (`hksv_recording.py`) matching the HAP spec / HAP-NodeJS (supported vs. selected encoding, list separators, per-attribute resolutions).
- **Camera wiring** — `CameraRecordingManagement`, `CameraOperatingMode`, `DataStreamTransportManagement`, an event-trigger `MotionSensor`, the recording-stream delegate hook. Streaming is set up only when video options are present, so a recording-only camera is possible.
- **SetupDataStreamTransport write-response** — the setter returns the encoded response (TCP port + accessory key salt) so the HAP layer delivers it in the write response (first user of the `wr` mechanism).
- **Integration helpers** — `fmp4_recording_packets(reader)` turns any fragmented-MP4 byte source (ffmpeg stdout, an aiohttp/go2rtc body, ...) into the init packet + fragments a delegate yields; `Camera.set_motion_detected(bool)` drives the trigger. A consumer supplies a keyframe-aligned fMP4 source and a motion signal rather than reimplementing framing/protocol.

## Robustness (asyncio / resources)

- Pre-handshake receive buffer capped + handshake timeout so an unauthenticated peer on the port can't make us buffer unboundedly; frames over a 1 MiB payload cap are rejected.
- Pending transport registrations bounded; malformed frames/messages and handler errors contained (bounded opack recursion, guarded dispatch); a stale `dataSend` close can't stop the current stream.
- Recording task + delegate torn down exactly once on close and/or connection loss, and the delegate's async generator is `aclose()`d so its subprocess is released deterministically.
- Fragment delivery honours transport write back-pressure (`pause`/`resume_writing`).

## Testing

- New unit tests: HDS transfer/chunking, connection-loss cleanup, double-stop, back-pressure, buffer/pending bounds, config round-trips (incl. multi-value), setup/eligibility, the write-response handshake, the fMP4 framer. Full suite green.
- `flake8` + `pylint` (repo configs) clean; `black`/`isort` applied.
- Manually verified end-to-end on real iOS 26.5 hardware (pair -> enable recording -> motion -> fMP4 fragments over HDS).

## Notes

- The shared `Version` characteristic (0x37) is narrowed to paired-read only (per spec; it previously also advertised events the accessory never emitted). Two `to_HAP` test expectations updated accordingly.
- Includes a small Python 3.14 compat guard for the removed `asyncio.SafeChildWatcher`.
- Happy to add a `CHANGELOG.md` entry if you'd like it in this PR — left it out so you can place it under the version/section you prefer.

### DIFF
--- a/pyhap/accessory_driver.py
+++ b/pyhap/accessory_driver.py
@@ -292,6 +292,9 @@ class AccessoryDriver:
         self.persist_file = os.path.expanduser(persist_file)
         self.encoder = encoder or AccessoryEncoder()
         self.topics = {}  # topic: set of (address, port) of subscribed clients
+        # Per-session HAP shared keys, keyed by client (address, port). Needed
+        # by transports derived from the session, e.g. HomeKit Data Stream.
+        self.session_shared_keys: Dict[Tuple[str, int], bytes] = {}
         self.loader = loader or Loader()
         self.aio_stop_event = None
         self.stop_event = threading.Event()
@@ -322,6 +325,9 @@ class AccessoryDriver:
             if (
                 threading.current_thread() is threading.main_thread()
                 and os.name != "nt"
+                # The child watcher API was removed in Python 3.14; there the
+                # default event loop reaps subprocesses without one.
+                and hasattr(asyncio, "SafeChildWatcher")
             ):
                 logger.debug("Setting child watcher")
                 watcher = asyncio.SafeChildWatcher()  # pylint: disable=deprecated-class
@@ -537,6 +543,7 @@ class AccessoryDriver:
         for topic in client_topics:
             self.async_subscribe_client_topic(client, topic, subscribe=False)
         self.prepared_writes.pop(client, None)
+        self.session_shared_keys.pop(client, None)
 
     def publish(self, data, sender_client_addr=None, immediate=False):
         """Publishes an event to the client.
--- a/pyhap/camera.py
+++ b/pyhap/camera.py
@@ -25,10 +25,10 @@ import struct
 import sys
 from uuid import UUID
 
-from pyhap import RESOURCE_DIR, tlv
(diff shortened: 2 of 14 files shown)


RaHehl:
Some context on where this fits:

This implements the **classic HKSV recording path** — HDS + `CameraRecordingManagement` + fragmented-MP4, the H.264 baseline every controller understands and the recording model current iOS uses. It works on today's iOS (verified live end-to-end) and stands on its own.

It's also intended as the **foundation for Apple's newer HomeKit Secure Video spec** — the [HomeKit Secure Video Open Source Compatibility Guide](https://developer.apple.com/download/files/HomeKit-Secure-Video-Open-Source-Compatibility-Guide.pdf) (HEVC / multi-tier RTP, WebRTC, CMAF-ingest recording). That path keeps the classic H.264 recording configuration as the baseline and layers HEVC tiers on top, so a working classic HKSV implementation is the prerequisite. Landing this first gives HKSV a complete, working baseline now, with the HEVC / multi-tier layer as a follow-up.

